PR #24, opened 2022-08-06
Spotlight

JiatengTao:
The spotlight function is ready for production. The actions are updated for three tools and one attack vector for now. The group of each action has been defined and user can search each page by title and group  name.


Huskehhh:
This is absolutely brilliant.

I made a minor change, extracting all `Routes` to the `RouteWrapper` component, and then the `Spotlight` component just converts those routes into a `SpotlightAction` on the fly.

Brilliant work @JiatengTao

Huskehhh (approved):

PR #33, opened 2022-08-16
Feature: John The Ripper + Arm64 Support + Program Icon

PurpleSn0w:
- Added basic implementation of John The Ripper.

- Added dependencies for Arm64 support

- Added program icon


install_dependencies.sh
```diff
@@ -23,19 +23,33 @@ sudo apt install -y mitmproxy \
...
+curl -fsSL https://deb.nodesource.com/setup_18.x | bash -
+exit
+E0F
 
 # Install openssl.
+echo "---------------------"
 echo "Installing openssl..."
+echo "---------------------"
 pushd /tmp
 wget http://nz2.archive.ubuntu.com/ubuntu/pool/main/o/openssl/libssl1.1_1.1.1f-1ubuntu2.16_amd64.deb
```
Huskehhh, line 54: Remove the installation of volta now that nodesource is the install method of node

install_dependencies.sh
```diff
@@ -23,19 +23,33 @@ sudo apt install -y mitmproxy \
         libjavascriptcoregtk-4.0-18 \
         libjavascriptcoregtk-4.0-dev \
         libwebkit2gtk-4.1 \
-        libwebkit2gtk-4.1-dev
+        libwebkit2gtk-4.1-dev \
+		cargo \
+		nodejs
```
Huskehhh, line 28: This needs to be done AFTER node source is installed

install_dependencies.sh
```diff
@@ -50,18 +64,24 @@ volta setup
 export VOLTA_HOME="$HOME/.volta"
 export PATH="$VOLTA_HOME/bin:$PATH"
 
+echo "---------------------------"
 echo "Installing node and yarn..."
+echo "---------------------------"
 # Install node and yarn.
 volta install node
 volta install yarn
```
Huskehhh, line 72: Install yarn via NPM instead now that node source is being used

src-tauri/Cargo.toml
```diff
@@ -17,7 +17,7 @@ tauri-build = { version = "1", features = [] }
 [dependencies]
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
-tauri = { version = "1", features = ["shell-execute", "shell-open", "shell-sidecar"] }
+tauri = { version = "1", features = ["fs-all", "path-all", "shell-all"] }
```
Huskehhh, line 20: Nice!

src/components/JohnTheRipper/JohnTheRipper.tsx
```diff
@@ -0,0 +1,99 @@
...
+                setOutput(result);
+            } catch (e: any) {
+                console.log(e);
+                setOutput(e);
+            }
+
+            try {
+                const args = [`--format=${values.fileType}`, "/tmp/hash.txt"];
+                const result = await CommandHelper.runCommand("john", args);
+                console.log(result);
```
Huskehhh, line 48: Remove this (and others similar)

Huskehhh (changes requested):
Mostly looks great! Just needs a few minor changes for the ARM64 support

src/components/JohnTheRipper/JohnTheRipper.tsx
PurpleSn0w: This is required. Unless there is a way to shorten the code.
In the if true statement
The first command find the hash and stores it. The second reads the hash and cracks the password.
In the if else statement
Reads the user inputed hash and cracks the password.

install_dependencies.sh
PurpleSn0w: Made some adjustments and pushed an updated install_dependencies.sh

src/components/JohnTheRipper/JohnTheRipper.tsx
Huskehhh: Sorry, meant the `console.log`

PurpleSn0w: Done

Huskehhh (approved):

PR #40, opened 2022-08-23
Hashcat

JiatengTao:
Basic functions in hashcat are implemented and ready for production


PurpleSn0w (changes requested):
Thanks for the submission. Could you please add the appropriate entry in the references page?

PR #58, opened 2022-09-10
Vulscan Tool Completed

tjmat1:
Added completed Vulscan Tool
Reference Added

### DIFF
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -39,7 +39,8 @@ sudo apt install -y mitmproxy \
         libwebkit2gtk-4.1-dev \
 		openjdk-11-jdk
 		cargo \
-		nodejs
+		nodejs \
+		dsniff
 
 echo "-----------------------------"
 echo "Installing yarn..."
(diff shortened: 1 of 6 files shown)


Huskehhh (changes requested):
The code looks great. I am however unable to execute any of the scripts, see below image.

image

tjmat1:
Hi Jordan,

The vulscan module requires manual installation into the scripts folder of the Nmap installation on the user's system (Instructions on https://github.com/scipag/vulscan). 

Could we add this as a requirement before the tool is used? Or did we need to add the installation into the project code itself (I wasn't sure if this was necessary?).

Thanks :)

tjmat1 (commented):
Please refer to comment above :)

Huskehhh:
> Hi Jordan,
> 
> The vulscan module requires manual installation into the scripts folder of the Nmap installation on the user's system (Instructions on https://github.com/scipag/vulscan).
> 
> Could we add this as a requirement before the tool is used? Or did we need to add the installation into the project code itself (I wasn't sure if this was necessary?).
> 
> Thanks :)

Good call. We should look at doing this in the `install_dependencies.sh` script

PR #74, opened 2022-09-24
Apache apisix rce

tjmat1:
-Added another attack vector CVE-2022-24112 Apache APISIX RCE
-tested successfully in virtual lab

exploit_six_output


PurpleSn0w (approved):
Approving based on your testing prior to submission.

PR #121, opened 2023-04-03
Userguide

whirlingbobo:
First pass at the UserGuide component. Functionality tested by importing into NmapTool. UserGuide parameters are two strings, a title for the tool it is being used in, and a description and link to a user guide for the tool. UserGuide returns a title with a small red question mark icon that uses HoverCards to reveal the description and link when the mouse is over the icon.


JiatengTao:
This PR is replaced by #123

PR #122, opened 2023-04-03
Dnsenum tool installed

camc375:
branch contains implemented DNSenumTool.ts file under the path src/components/DNSenumTool. Tool should operate correctly with the GUI, return enumerated dns information on target domain. option available to increase or decrease the thread count of the attack (can speed up attack, at the cost of being more visible to security programs). changes also made in RouteWrapper.tsx and tauri.conf.json to allow the tool to run.


JiatengTao:
This pr is replace by #124

PR #127, opened 2023-04-05
Pk contri

PushkarGoel:
I have successfully added cewl tool, The foremost tool is also working but there is some issue with paths so working on that atm


JiatengTao:
Have you edited the yarn.lock file manually?

PushkarGoel:
No, I didn't but there was a conflict so I just merged my version with
older version using git.However I missed one merge conflict marker and
that's why I think the build failed.

I corrected the file and then commit it but the build did not start on that
commit. I am a little confused about that

JiatengTao:
It still has the same error, I think it will work if you can clone the main branch to a new local branch and paste the file you edited for the cewl tool?

JiatengTao:
It won't because of the newly merged branch, the yarn.lock throughs error only because the dependencies packages

JiatengTao:
You have 427 files changed now, and I can see you changed your yarn.lock in #337f887

PR #177, opened 2023-05-04
Arjuntool1 - Arjun is a command line tool specifically designed to look for hidden HTTP parameters.

laasya1999:


JiatengTao:
Hi, the routewrapper and tauri.config.json file also needed to be updated.

PR #210, opened 2023-05-17
FTP tool

GerardAraullo:
Added FTP tool for convenience.
- The tool opens up a separate interactive terminal where the user can use FTP


JiatengTao:
One small issue, FTPterminal.sh should be uploaded to the exploit folder

PR #225, opened 2023-05-20
Eyewitness tool is used to take screenshots of target web Urls

AnuragVerma26:


GerardAraullo:
Based on the error you forgot to run the prettier command on routewrapper.tsx

PR #232, opened 2023-05-20
The harvester tool Gui

Sharfyy:


JiatengTao:
1. It's okay if you want to move this PR into integrating BeautifulSoup into the toolkit, but for Harvester you should trigger it from the Kali.
2. You need to register it in the Routewrapper to make the GUI show up.

PR #233, opened 2023-05-20
3 tools

Sharfyy:


JiatengTao:
1. The msfvenom.py length check is not correct
2. The gobuster.py will receive gobuster no such file or dir from subproccess
3. The trigger path for python files should be '/usr/share/ddt/'
The Harvest works well and the solution is unique. However, for project consistency, need discussion with the company director before any merging.

PR #275, opened 2023-08-06
Update README.md

whirlingbobo:
Updated install instructions to reflect current process


whirlingbobo:
Need to run npx prettier on local machine to pass style test

PR #287, opened 2023-08-11
Add fcrackzip tool

kbachhal:


whirlingbobo:
Need to add tauri.conf.json

kbachhal:
added

PR #291, opened 2023-08-12
Update NmapTool.tsx

Blitzle8264:
Added "Advanced Mode" toggle including following features: • Exclusion ranges (for when removing an IP from a selected range) • Verbose mode (moved from general screen)
• IPv6 Address mode

Renamed "All" scan option to "Default" as previously version of "All" was a misnamed Aggressive scan. Will look at working on adding an "All" scan option


whirlingbobo:
Failed checks, superseded by @292

PR #308, opened 2023-08-31
dmitry audit

hell0karl:
added advanced mode, optimized options in GUI, added cancel button


whirlingbobo:
Redundant. PR #309 is preferred to include Dmitry and Crackmapexec updates.

PR #429, opened 2023-11-18
Fix issues with DNSMap tool as per the PTGUI Final Report

v-xl:
Fixed the following:

1. Spelling errors in the tooltip text
2. When selecting an external wordlist, the DDT GUI uses the incorrect command -S instead of -w.
3. When selecting results output to a CSV file, the DDT GUI uses the incorrect command -s instead of -c.
4. When selecting an IP address to ignore, the DDT GUI uses the incorrect command -t instead of -i.

### DIFF
--- a/src/components/DNSMap/DNSMap.tsx
+++ b/src/components/DNSMap/DNSMap.tsx
@@ -10,15 +10,15 @@ import { LoadingOverlayAndCancelButton } from "../OverlayAndCancelButton/Overlay
 const title = "DNS Mapping for Subdomains (DNSMap)";
 const description_userguide =
     "DNSMap scans a domain for common subdomains using a built-in or an external wordlist (if specified using -w option). " +
-    "The internal wordlist has around 1000 words in English and Spanish as ns1, firewall servicios and smtp. " +
+    "The internal wordlist has around 1000 words in English and Spanish as ns1, firewall services and smtp. " +
     "So it will be possible search for smtp.example.com inside example.com automatically.\n\nInformation on the tool " +
     "can be found at: https://www.kali.org/tools/dnsmap/\n\n" +
     "Step 1: Enter a valid domain to be mapped.\n" +
     "       Eg: google.com\n\n" +
     "Step 2: Enter a delay between requests. Default is 10 (milliseconds). Can be left blank.\n" +
     "       Eg: 10\n\n" +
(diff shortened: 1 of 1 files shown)


v-xl:
Just adding another note for my second commit from yesterday:

I had found that the command line args for DNSMap weren't working properly, and were either being ignored or causing errors upon the command execution. 

For example, when specifying a wordlist, it would not be used and when specifying a results .csv file it would also be ignored.

After doing some testing and making some changes, I had been able to get the tool functioning as intended. 

As can be seen in the commit, I've had to split into 2 separate args for each of the following:
- wordlistPath
- csvResultFile
- ipsToIgnore

src/components/DNSMap/DNSMap.tsx
```diff
@@ -110,15 +110,15 @@ const DNSMap = () => {
         const args = [`${values.domain}`, "-d", `${values.delay}`];
 
         if (values.wordlistPath) {
-            args.push(`-S ${values.wordlistPath}`);
+            args.push(`-w`, `${values.wordlistPath}`);
```
Sakada-lim, line 113: This code work fine but i don't see why you would need to wrap your `values.wordlistPath` in a `''` as in the form the value is already define as a string. Same goes with the other if statement.

src/components/DNSMap/DNSMap.tsx
```diff
@@ -110,15 +110,15 @@ const DNSMap = () => {
...
         }
 
         if (values.csvResultsFile) {
-            args.push(`-s ${values.csvResultsFile}`);
+            args.push(`-c`, `${values.csvResultsFile}`);
```
Sakada-lim, line 117: ### Suggestion for improvment
 For now the if statement is completely fine but if you want to improve the scalability of the code for future work you could look into expression such as ternary operator or a a better way of doing if statement.

Sakada-lim (commented):
Overall the code look fine but can be improve if you have time:
### suggestion for improvement
improve the if statement for scalability

Sakada-lim (approved):

Phaellynn (approved):
Passed all functional and useability tests as per Gangas Marx
Screenshot 2023-11-26 210637

gangamarx (approved):
Passed all functional and useability tests. 
Test Execution Results - DNSMapping

PR #430, opened 2023-11-18
John the ripper improvement

zkkok:
# Changes

**- Zip2John/Rar2John functionality**: Added functionality to extract and crack passwords protecting encrypted zip and rar files.
**- Raw Files**: added "raw" option under File Type to make sure text files are being cracked properly
**- Syntax Fixes**: Fixed syntax for different scenarios depending on inputs
**- Modified User Guide**: Made minor changes to user guide to provide more information after the changes
**- Cracking Modes**: Added Dictionary, Single and Incremental cracking modes and configure their follow up options

### DIFF
--- a/src/components/JohnTheRipper/JohnTheRipper.tsx
+++ b/src/components/JohnTheRipper/JohnTheRipper.tsx
@@ -105,30 +105,32 @@ const JohnTheRipper = () => {
         if (values.fileType === "raw") {
             var args: string[];
 
-            //change argument according to mode selected
-            if (selectedModeOption === "dictionary") {
-                args = [`--wordlist=${values.wordlist}`, `${values.filePath}`];
-            } else if (selectedModeOption === "incremental") {
-                args = [`-incremental:${values.incrementorder}`, `${values.filePath}`];
-            } else {
-                args = [`--single`, `${values.filePath}`];
-            }
-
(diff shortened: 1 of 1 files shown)


Phaellynn (approved):
JohnTheRipper tool passed functionality tests for:
- Raw option added to file types
- Raw hash files can be cracked
- Password-protected zip files can be cracked
- Password-protected rar files can be cracked
Screenshot 2023-11-22 131826

src/components/JohnTheRipper/JohnTheRipper.tsx
```diff
@@ -71,32 +103,74 @@ const JohnTheRipper = () => {
     const onSubmit = async (values: FormValuesType) => {
         setLoading(true);
 
-        //if hash is not specified
-        if (values.hash.length === 0) {
-            const args = [`${values.filePath}`];
+        //if hash is stored in a textfile
```
Sakada-lim, line 106: The onSubmit logic is Great. If you would like to improve on this, you can consider extracting the args into its own function that return value depend on the parameter input (functional approach). This would make the code a lot more scalable and less singleton.

Sakada-lim (approved):
Overall code is great but maybe consider extraction some function into it's own logic to make the code more scalable and less singleton like.

PR #432, opened 2023-11-26
Fixed GoBuster tool functionaility + implemented SaveOutputToTextFile_v2

v-xl:
In the Tauri config, for the gobuster command, the ‘G’ was capitalised which is the cause of the issue, since the command itself should be executed in all lower case. This has been modified and tested successfully.

A bug I found with this tool, the existing Output file textbox didn't do anything, and no Output file was created upon running the tool. 

I've modified the code to remove the existing Output textbox, and I've instead implemented the SaveOutputToTextFile_v2 function as per the documentation in Teams. After modifying the code and implementing the SaveOutputToTextFile_v2 function all is working as intended and the output file will be created.


src/components/GobusterTool/Gobuster.tsx
```diff
@@ -52,8 +104,8 @@ const GoBusterTool = () => {
                 {UserGuide(title, description_userguide)}
                 <TextInput label={"Target URL"} required {...form.getInputProps("url")} />
                 <TextInput label={"Wordlist File"} required {...form.getInputProps("wordlist")} />
-                <TextInput label={"Output File (optional)"} {...form.getInputProps("outputFile")} />
                 <Button type={"submit"}>Start Brute-Forcing</Button>
+                {SaveOutputToTextFile_v2(output, allowSave, hasSaved, handleSaveComplete)}
```
Sakada-lim, line 108: This is not a bug. `-o` is a feature that allow the user to save the output result to a file name. However, with that being said, it is very similar to `{SaveOutputTotextFile_v2}`. The difference is `{SaveOutputTotextFile_v2}` save the whole command include all the print value during the execution but `-o` only print the result.

Sakada-lim (approved):
Overall the code is good but more feature should be added for gobuster. 
### idea for improvement

1. Add more feature for gobuster
2. implement the `-o` feature alongside the `{Saveoutputtofile_v2}`
3. make the onsubmit more scalable

gangamarx:
gobuster works well in CLI mode but do not work at all in GUI. Please looks at the screen shots
gobuster works in CLI mode
gobuster dont work in gui for any URL

gangamarx (changes requested):
gobuster works in CLI mode but not in GUI mode.
gobuster works in CLI mode
gobuster dont work in gui for any URL

v-xl:
Thank you, this was due to the Tauri config. It looks like I've forgotten to commit the updated file.
I've now fixed this.

gangamarx:
I will mark this task as complete.

Phaellynn (approved):
GoBuster has passed all functionality tests as reviewed by Ganga Marx
Tool now scans as expected without errors.
image

gangamarx (approved):
The tool now works as expected without any errors.

Issue #414, opened 2023-11-14
[Bug]: Metagoofil


Sakada-lim:
### Name of Bug

Files are not being download

### What Happened?

Metagoofil, a metadata extraction tool, is not functioning correctly within the DDT GUI environment.

### Steps to Reproduce

refer to the report

PR #437, opened 2023-11-29
Metagoofil improvement

zkkok:
# Changes

- **File Download:** Added the functionality to download files that are found
- **Removed Unnecessary Requirements:** Removed required fields for search and file limits
- **Minor Spelling Errors:** Fixed some typos in the user guides
- **Fixed Syntax:** The code was previously running on "cewl", changed to "metagoofil" accordingly

Link to issue #414

### DIFF
--- a/src/components/metagoofil/metagoofil.tsx
+++ b/src/components/metagoofil/metagoofil.tsx
@@ -23,7 +23,7 @@ const description_userguide =
     "Step 4: Enter the file type name to be extracted.\n" +
     "       Eg: pdf\n\n" +
     "Step 5: Click scan to commence the Metagoofil operation.\n\n" +
-    "Step 6: View the Output block below to view the results of the tools execution.";
+    "Step 6: View the Output block below to view the results of the tool's execution.";
 
 interface FormValues {
     webname: string;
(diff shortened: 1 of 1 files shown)


src/components/metagoofil/metagoofil.tsx
```diff
@@ -85,15 +88,21 @@ export function Metagoofil() {
         setLoading(true);
 
         const args = [`-d`, `${values.webname}`, `-t`, `${values.filetype}`];
-
+        //number of searches made
```
Sakada-lim, line 91: if statement is fine but maybe hard to scale later down the line. Maybe consider a new logic that can scale for the future.

Sakada-lim (approved):
code is good. maybe consider scalability for future update

zkkok:
@gangamarx I think you are performing the test on the old code 
image
I can see "cewl" on the output but the new code has switched to "metagoofil"

gangamarx:
Thank you zkkok, yes you are correct, and I was using the old code. 
Now I have updated the code and checked with all parameters as stated and it works well without errors both in CLI and GUI mode (both standard and manual) as per below images. 

GUI Works -std
GUI Works-manual

Phaellynn (approved):
Metagoofil has passed all functionality tests as tested by Ganga Marx
- Grammatical Errors in ToolTip are fixed
- Document Download Error is now fixed
Screenshot 2023-12-13 133419

gangamarx (approved):
The tool now works as expected without any errors.

PR #440, opened 2023-12-02
Improvment for DirbTool

Abhijeet-97:


Phaellynn (commented):
Dirb - Submit Button added to Std Mode - Issue Fixed
Screenshot submit button added

**NB: None of the other issues are fixed. Only this one.** But, if this was the only issue the developer was expected to fix, then it is fixed.

Sakada-lim (approved):
adding just the button for the execution. The execution of the code still doesn't function as expected

PR #444, opened 2023-12-02
Userguide foremost

YashKanodia:


Phaellynn (changes requested):
Foremost - Issues below are not fixed
1. User can still select both Quiet & Verbose mode simultaneously. Issue Not fixed
Screenshot quiet verbose

2. When selecting 'Write All Headers' option, the tool issues the -d option instead of -a in the command.
Screenshot option -a

PR #445, opened 2023-12-02
Userguide frackzip

YashKanodia:


Phaellynn (changes requested):
Fcrackzip - 
1. When 'Show Cracked Password' is selected user can no longer select a character set (the Lowercase/Uppercase/Numeric) boxes become inactive. - Issue not Fixed

PR #448, opened 2023-12-09
Foremost

YashKanodia:
Foremost - Issues below are fixed

1. User can still select both Quiet & Verbose mode simultaneously. 
1. When selecting 'Write All Headers' option, the tool issues the -d option instead of -a in the command.


Phaellynn (changes requested):
@YashKanodia 
Foremost Test Review as follows:
- Write All Headers Issue has been fixed with this change.
-  When testing the Quiet/Verbose issue it produced an error. When selecting the Quiet or the Verbose checkboxes more than once, the application crashes with a white screen:
Screenshot 2023-12-09 184557sm

Please see the type error generated in your code post amendments. 
Screenshot 2023-12-10 124909
Screenshot 2023-12-10 124921

This is a type error. Please see some common fixes:
1. Check variable types.
2. Verify function arguments match parameters.
3. Use type assertions when certain about a type.
4. Implement type guards for type checking.

PR #449, opened 2023-12-09
Userguide arjuntool

Abhijeet-97:
As per the previous reviews for the changes I have removed the First field for optional saving the output field. Now the output is saving in one field.


Phaellynn (approved):
Branch: userguideArjuntool
Tool passes functionality test:

Tool now has only one field for writing a file name for saving output to disk.
Tool successfully saves the output to a file on the hard disk.
Screenshot 2023-12-09 162516

Sakada-lim (approved):
There are conflicting file that were not suppose to be work on by the developer. I will merge and manually revert the chance for some of the file.

Issue #417, opened 2023-11-14
[Bug]: Arp Spoofing


Sakada-lim:
### Name of Bug

Tool doesn't work

### What Happened?

During a test for ARP Spoofing, which is a Man in the Middle (MitM) attack technique, the expected ARP spoofing was not visible in Wireshark. This tool typically sends forged ARP responses to mislead at least two devices on a network, causing them to communicate with the attacker's MAC address instead of the true host.

### Steps to Reproduce

Refer to the Report

PR #450, opened 2023-12-09
Arp spoof improvement

zkkok:
# Changes

- **Functionality:** ArpSpoof now successfully spoofs the arp table
- **Cancel Function:** The stop button is now able to properly terminate the process and avoid them from running on the background.
- **Updated interface:** Original interface only has two fields and one alert to tell the user if the process is running. Added output field and ability to save the output file to provide a more verbose output

Link to issue #417


Phaellynn (approved):
ARPSpoof Tool passed all functionality test:
- Spoofing executes successfully creating a MiTM scenario
- Output saved to file on hard disk
Screenshot 2023-12-10 163155

Sakada-lim (approved):
code is great maybe just instead of handle1 and 2 come up with a better name but perfect fine.

PR #451, opened 2023-12-09
Fcrackzip

YashKanodia:
Fcrackzip - The below issue is fixed 

When 'Show Cracked Password' is selected user can no longer select a character set (the Lowercase/Uppercase/Numeric) boxes become inactive. 

gmail

Here, in the screenshot I have checked "Use UpperCase" and "Use Numeric" and when checked "Show Cracked Password" the boxes are still active.


Phaellynn (dismissed):
This issue is only HALF fixed. So issue is not fixed.
- When selecting the Charset List BEFORE selecting the 'Show cracked password' checkbox, the selection remains active, but still 'greyed out' and you cannot update the Charset List.
- You cannot select the Charset List AFTER selecting the 'Show cracked password' checkbox.
Screenshot 2023-12-09 180735

- The same issue is not present when selecting the 'Verbose mode' checkbox.
Screenshot 2023-12-09 180702

- There also appears to be an extra issue that wasn't present before. There are now TWO 'Clear Output' buttons instead of just the one. Please remove one of them.
Screenshot 2023-12-09 181057

YashKanodia:
As discussed with Liz, there's a pending discussion with Sakada and Bradley on the above issues.

bradleybogg (changes requested):

src/components/Fcrackzip/Fcrackzip.tsx
Sakada-lim, line 1: admin have personally adjust this pull request and make some change in order to make the code work

Sakada-lim (approved):
Adjusted and tested by Admin

PR #455, opened 2023-12-10
fix: changed from advanced to custom input

BlokeyAU2:
**Issue:** 
Issue discussed with @Phaellynn, original WPScan advanced mode GUI was too complicated and didn't work correctly. Rectified this by implementing a custom input field. 

**Resolution**
In the commit (fix: https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/commit/b6b2c43702b469aca1108202b62bf0d3d4e46fab) The GUI interface was changed to a more appropriate and functional version. 

**Changes made**
Modified the GUI and added a custom input field for the user to put their own flags in. Changed the Tooltip to reflect the new custom input field. 

**Testing**
Awaiting testing from testing team

- Tested all the advanced mode settings, and deduced they were broken
- Deleted the original advanced mode and created a custom input field
- Replaced all the switch statements in WPScan with a new textinput field
- Rewrote the tooltip for WPscan to reflect changes


Phaellynn (changes requested):
WPScan:
- The basic scan now works as expected
- The GUI looks much nicer and easier to use
- Unfortunately, the Custom Flags field isn't working as expected. It adds the correct option to the command in the output display, but this doesn't seem to be transferring to the program, as the option either isn't included in the scan or produces an error.
Screenshot 2023-12-11 171640

Issue #409, opened 2023-11-14
[Feature Request]: Airbase-ng


Sakada-lim:
### Name of Feature

More feature require

### Categories

implementation

### Description of Feature

need more feature implemented

PR #457, opened 2023-12-12
Airbase ng improvement

zkkok:
# Changes
- Advanced mode: Added some fields that could be common when using the tool
- Custom mode: Added a field for custom configurations to provide more flexibility
- Manual Stop Button: Added a manual stop button since the tool is running on pkexec

Link to issue #409


Sakada-lim:
The code is great with no issue insight. You can also improve further on the pkexec as there are some tools that will be using pkexec in the future for privileged issues so you can create a new tools for deal with the the pkexec threading issue in general so all tools future can import and use the stop button.

PR #458, opened 2023-12-13
Nbtscan

bradleybogg:
PR requested onbehalf of Renee Mateo.

Renee to amend description.


majore15:
Nbtcan should only consist of "One option only in Advanced Mode" and "Fix Advanced Mode processing syntax"

Phaellynn (changes requested):
This works really well. Original issue has been fixed.
I would just recommend two small changes:
1. The 'Subnet' and 'Range to Scan' fields accept both values. Can we please change the Label for both to display as 'Subnet or Range to Scan' ?
2. The Dump Packets Mode checkbox (in Advanced Mode) does nothing. Can we either remove it or have it add the -d option?

bradleybogg (commented):
Please remove these files from your PR:

- [x] Yarn.lock
- [x] install-state.gz

In future please also only make one component change per branch/PR

Phaellynn (commented):
NBTScan passed all functionality tests:
- Subnet/Range-to-Scan Error fixed
- Dump Packets Error Fixed
Screenshot 2023-12-14 222612
Screenshot 2023-12-14 222549

majore15:
@bradleybogg 
Completed pull merge.  Please review.  Thanks.

src/components/NbtscanTool/NbtscanTool.tsx
```diff
@@ -141,7 +145,6 @@ const NbtscanTool = () => {
         setAllowSave(false);
     }, [setOutput]);
 
-    //define user interface of the tool
     return (
         <form onSubmit={form.onSubmit(onSubmit)}>
```
Sakada-lim, line 149: This build fail because in your interface you have declare promise such as DumpPacket but in your form you have not include it in a value that we pass thru. I see that you have created a usestate instead for your dumppacket which is fine but you will need to remove the interface promise in order to pass the build.

src/components/NbtscanTool/NbtscanTool.tsx
```diff
@@ -41,11 +41,11 @@ const NbtscanTool = () => {
     const [pid, setPid] = useState("");
     const [allowSave, setAllowSave] = useState(false);
     const [hasSaved, setHasSaved] = useState(false);
+    const [subnet, setsubnet] = useState("");
```
Sakada-lim, line 44: you have created a subnet usestate() but never use them.

src/components/NbtscanTool/NbtscanTool.tsx
Sakada-lim, line 1: The code is still buggy as when advance mode is toggle on the subnet button disappear however it's input still presist.

Sakada-lim (changes requested):
You will need to change the interface promise to pass the build test.

Sakada-lim (approved):

PR #460, opened 2023-12-13
Foremsot

YashKanodia:
The previous version had below issues:
When testing the Quiet/Verbose issue it produced an error. When selecting the Quiet or the Verbose checkboxes more than once, the application crashes with a white screen.

Was checking with the issue but for now, we can select either of the modes but when one of the selection is same and try selecting the other the application. 
These below were checked to mitigate the issue:
1. Check variable types.
2. Verify function arguments match parameters.
3. Use type assertions when certain about a type.
4. Implement type guards for type checking.


Phaellynn (changes requested):
@YashKanodia 
This is getting better. The original issue has been addressed. 
- The user can only select one of the checkboxes (either Quiet or Verbose), but not both. That Issue has been Fixed.

A new issue has been created:
- When the user selects/deselects the Quiet or Verbose Checkboxes 3 times the DDT application displays an all white screen with only the Title Header visible.
Screenshot 2023-12-14 060315

When the element is inspected the following error messages are displayed:
Screenshot 2023-12-14 060015

bradleybogg (changes requested):
Hi @YashKanodia,

Please separate these to their respective branches. We can address them individually.

PR #461, opened 2023-12-13
Userguide dirb tool

Abhijeet-97:
I have Set the  verbose mode as optional to simplify result analysis.

### DIFF
--- a/src/components/Arjuntool/Arjuntool.tsx
+++ b/src/components/Arjuntool/Arjuntool.tsx
@@ -113,7 +113,7 @@ export function Arjuntool() {
                 <TextInput label={"URL"} required {...form.getInputProps("url")} />
                 <TextInput
                     label={"Optional Json output file: provide file name if required"}
-                    {...form.getInputProps("o")}
+                    {...form.getInputProps("output_filename")}
                 />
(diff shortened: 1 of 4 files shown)


Phaellynn (changes requested):
@Abhijeet-97 
I can see that you have added the -v option to the advanced mode, which is great.
The real issue is that the DDT display prints everything that appears on the screen when the tool is run (giving the appearance of verbose mode, even when it has not been requested). 
When the tool runs from the CLI, the screen overwrites and replaces the subdomains it tests and only leaves the matching ones it finds.
Screenshot 2023-12-14 065028

The only way to duplicate this is to set the Simple Scan mode to run in Silent Mode. This can be done by setting the -S option. Can you please program it so that Simple Mode always runs with the Silent option (-S)?

Abhijeet-97:
I have added the Silent option in simple mode. can you please review it to see if the issue is fixed now. Thank you

Phaellynn (approved):
@Abhijeet-97 
Dirb Tool has passed functionality test for Verbose Mode Error
- Verbose mode is off for Simple/Standard Mode
- Verbose Mode is on for Advanced Mode
Screenshot 2023-12-14 121943

PR #466, opened 2023-12-14
Fix <Unknown Error> when using port scan

majore15:


bradleybogg (changes requested):
How to fix
Run: yarn run style in the DDT terminal

You should see the files that has the styling problem

Run: npx prettier --write <filepath> to fix it

bradleybogg (commented):
Ran prettier. CI now passing.

Phaellynn (approved):
Netcat passed functionality tests for:
- Port Range scan issue
image
NB: Send/Receive File Error is still outstanding.

Issue #419, opened 2023-11-14
[Bug]: Gyothion


Sakada-lim:
### Name of Bug

tools doesn't work

### What Happened?

The installation process for the GyoiThon tool encounters an error due to the `requirements.txt` file being in an unexpected directory.

### Steps to Reproduce

refer to the report

slpysiren:
The paths used in the Gyoithon component were absolute and therefore wouldn't work unless you had the exact folder structure that whoever wrote them did. I changed the paths to be relative to the project so they will be correct for any install from the repo.

Useful note: any commands run with the `CommandHelper` util execute from `src-tauri`, so if a component uses that util and you want to point to a file or directory, it will need to be relative to `src-tauri` e.g. `./exploits/GyoiThon/gyoithon.py`

bradleybogg:
From the T2 2023 report:
Users need to install dependencies for the Gyoithon tool to work. This process results in an error as the requirements.txt file is not in the directory the installation script expects.

![image](https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/assets/62839826/51efff7b-01eb-4f22-9dcc-20d8235aa386)

Installation of dependencies step was continued but again resulted in an error from a subprocess. As a result, the program halted the installation of dependencies. The Gyoithon tool could not be installed, subsequently could not run, and the test could continue.

![image](https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/assets/62839826/9c29aa58-8e09-4a75-973f-fea9866f5071)

slpysiren:
> The paths used in the Gyoithon component were absolute and therefore wouldn't work unless you had the exact folder structure that whoever wrote them did. I changed the paths to be relative to the project so they will be correct for any install from the repo.
> 
> Useful note: any commands run with the `CommandHelper` util execute from `src-tauri`, so if a component uses that util and you want to point to a file or directory, it will need to be relative to `src-tauri` e.g. `./exploits/GyoiThon/gyoithon.py`

I was incorrect in my assumption of the absolute path. The absolute path is correct if the exploits are installed as intended using the `install_exploits.sh` script located in `/install-update-media`. This alone fixes the bug as described in the initial issue.

slpysiren:
_Accidental closure, apologies._

[...]

PR #484, opened 2024-03-11
Fixed issue with GyoiThon paths

slpysiren:
Fixed issue referenced in #419 by changing paths in the GyoiThon component from absolute to relative. This has unearthed further bugs in this component that I will create new issues for soon.

### DIFF
--- a/src/components/Gyoithon/Gyoithon.tsx
+++ b/src/components/Gyoithon/Gyoithon.tsx
@@ -48,7 +48,7 @@ const Gyoithon = () => {
 
     const Install = async () => {
         setLoading(true);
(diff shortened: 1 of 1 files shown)


bradleybogg (changes requested):
Hi @slpysiren, 

Thanks for your first submission. 
Unfortunately, the static file path is deliberate. This avoids the installation location as a factor for runtime. 
It also allows the scripts to visible outside of the package / executable application. 

To understand this, I would recommend you install a packaged version of DDT and then try and access the scripts - you'll find they're not visible. 

The install-exploits.sh is run on installation of DDT (or such will be subject of future implementation) to make this happen. Refer to the install-media-folder to find it.

I have added more information to the issue for reference.

Cheers

bradley

PR #494, opened 2024-03-18
419 gyoithon dependencies fix

slpysiren:
Added extra dependencies to requirements.txt of GyoiThon allowing the tool to run, fixing #419 . The tool doesn't function fully but this is out of scope for the initial bug.


JiatengTao (approved):
Appreciate it, thanks

Issue #505, opened 2024-03-26
[Feature Request]: Give the option of stability for Arjun


michaeljpigott:
### Name of Feature

Stability option for Arjun

### Priority

Low

### Categories

Functionality

### Description of Feature

Arjun finds query parameters for URL endpoints. However, if it finds that a target is misbehaving, then it will suggest the --stable switch. It appears not to find parameters when the target is misbehaving: 

![arjun-stability](https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/assets/35479648/7f86f997-9839-4acd-b2b6-c9a1e153dc35)

This feature will add a stability checkbox to the Arjun tool, which means Arjun will prefer stability over speed. I have checked in Kali and adding the --stable option appears to get better results:

![kali-arjun-test](https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/assets/35479648/597ff47a-2d9a-4d02-99fc-e147cab5358f)

PR #525, opened 2024-04-06
505 feature request give the option of stability for arjun

michaeljpigott:
### Description

This pull request addresses issue #505, adding the stability switch to the Arjun tool. 

### Changes made 

* Imported the switch module into Arjuntool.tsx, so the tool can access the [switch feature](https://mantine.dev/core/switch/) from Mantine. 
* Added stability as a new boolean value of the FormValues object. 
* Set the initial value of the stability switch as false. 
* Added an if statement that pushes the option --stable to the args variable if the stability switch is turned on. 
* Added information to the tooltip that explains the stability option. 
* Updated the name of the tool from "Arjuntool" to "Arjun" in RouteWrapper.tsx. The tool is [called Arjun](https://www.kali.org/tools/arjun/) rather than Arjuntool. 
* Ran Prettier over Arjuntool.tsx to fix up style errors. 

### Testing

I ran the tool over Metasploitable's MyPhpAdmin and it ran successfully in both stability and non-stability modes. 

### Note

You will notice in the code that I originally changed the name of the Arjuntool component to Arjun, but I then changed it back. I changed my mind because I wanted the name of the component to match the name of the tsx file.

### DIFF
--- a/src/components/Arjuntool/Arjuntool.tsx
+++ b/src/components/Arjuntool/Arjuntool.tsx
@@ -14,8 +14,9 @@ const description_userguide =
     "It is a multi-threaded application, can handle rate limits and supports GET,POST,XML and JSON methods. " +
     " \n\nKali's Arjun Information Page: https://www.kali.org/tools/arjun/ \n\nHow to use Arjun:\n\nStep 1: Enter a valid URL.\n" +
     "       E.g. https://www.deakin.edu.au\n\nStep 2: Enter an Optional Json Output filename.\n        E.g. arjunoutput " +
-    "\n\nStep 3: Click the scan option to commence scan. " +
-    "\n\nStep 4: View the Output block below to view the results of the tool's execution.";
+    "\n\nStep 3: Turn on the stability function if you want Arjun to prioritise stability over speed (can take a long time, so get a coffee if using this functionality)." +
+    "\n\nStep 4: Click the scan option to commence scan. " +
+    "\n\nStep 5: View the Output block below to view the results of the tool's execution.";
 
 interface FormValues {
     url: string;
--- a/src/components/Arjuntool/Arjuntool.tsx
+++ b/src/components/Arjuntool/Arjuntool.tsx
@@ -87,7 +87,7 @@ export function Arjuntool() {
         const args = ["-u", values.url];
 
         // Conditional. If the user has specified stability, add the --stable option to the command.
-        if (values.stability == true) {
+        if (values.stability) {
             args.push("--stable");
         }
 
--- a/src/components/Arjuntool/Arjuntool.tsx
+++ b/src/components/Arjuntool/Arjuntool.tsx
@@ -24,7 +24,7 @@ interface FormValues {
     stability: boolean;
 }
 
-export function Arjun() {
+export function Arjuntool() {
     const [loading, setLoading] = useState(false);
     const [output, setOutput] = useState("");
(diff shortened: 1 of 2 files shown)


src/components/Arjuntool/Arjuntool.tsx
```diff
@@ -14,12 +14,14 @@ const description_userguide =
     "It is a multi-threaded application, can handle rate limits and supports GET,POST,XML and JSON methods. " +
     " \n\nKali's Arjun Information Page: https://www.kali.org/tools/arjun/ \n\nHow to use Arjun:\n\nStep 1: Enter a valid URL.\n" +
     "       E.g. https://www.deakin.edu.au\n\nStep 2: Enter an Optional Json Output filename.\n        E.g. arjunoutput " +
-    "\n\nStep 3: Click the scan option to commence scan. " +
-    "\n\nStep 4: View the Output block below to view the results of the tool's execution.";
+    "\n\nStep 3: Turn on the stability function if you want Arjun to prioritise stability over speed (can take a long time, so get a coffee if using this functionality)." +
+    "\n\nStep 4: Click the scan option to commence scan. " +
```
bradleybogg, line 18: Up to you I would have this read "to commence scanning".

src/components/Arjuntool/Arjuntool.tsx
```diff
@@ -14,12 +14,14 @@ const description_userguide =
...
+    "\n\nStep 5: View the Output block below to view the results of the tool's execution.";
```
bradleybogg, line 19: - [ ] Remove unnecessary capitalization.

src/components/Arjuntool/Arjuntool.tsx
```diff
@@ -83,6 +86,11 @@ export function Arjuntool() {
 
         const args = ["-u", values.url];
 
+        // Conditional. If the user has specified stability, add the --stable option to the command.
+        if (values.stability == true) {
```
bradleybogg, line 90: Hi @michaeljpigott,

This can be simplified. When conducting a conditional statement such as an if statement the argument inside () is used to check its truthiness of the expression.

Given that if the values.stability contains an argument it will be true the == true is redundant and can be removed.

- [ ] Remove == true and test.

Thank you!

bradleybogg (changes requested):
Hi @michaeljpigott,

Thanks for you hard work here.
I've recommended some small changes. 

Kind Regards,

Bradley

src/components/Arjuntool/Arjuntool.tsx
michaeljpigott: Yes, sounds good to me. I will change it.

src/components/Arjuntool/Arjuntool.tsx
michaeljpigott: Have done this. I also went through the tooltip again and made some slight changes to improve readability.

src/components/Arjuntool/Arjuntool.tsx
michaeljpigott: Tested and confirmed that it works without the == true.

src/components/Arjuntool/Arjuntool.tsx
michaeljpigott: All done.

bradleybogg (approved):

bradleybogg:
Thank you @michaeljpigott merged.

PR #527, opened 2024-04-08
Upgrade Mantine Version

JiatengTao:


JiatengTao:
Some features are deprecated in new mantine, need to be changed before upgrade

michaeljpigott:
Pull request no longer active.

PR #550, opened 2024-04-15
Fix for theHarvester Sources

isaiahspokes:
Updated List of sources in dropdown menu to remove deprecated sources and add new sources.
Added default option with no value to fix issue with the source value not being passed to the command line if source was not changed from default.


src/components/theharvester/theharvester.tsx
bradleybogg, line 1: Thanks for updating these. However, there is a better implementation.

Please use a react state hook to maintain tracking over the selected with a function selection of the array. 
A good example of this is in Netcat. It removes all of these lines that complicate without reason.

Thank you!

bradleybogg (changes requested):

bradleybogg (changes requested):
Hey @isaiahspokes 

This looks better. Only a slight change needed. Requires application testing and functionality test.

Thank you

bradleybogg:
@isaiahspokes Good work.

PR #638, opened 2024-04-28
507 feature request dig tool

SUA-321:
This DigTool component is a DNS lookup tool implemented in React.js.
It allows users to enter a domain name, performs a DNS lookup using the Google
Public DNS resolver, and displays the result or any errors that occur during
the process.

The component maintains state for the domain name, DNS lookup result, and error
messages. When the user clicks the "Perform Lookup" button, it triggers the
handleLookup function, which retrieves DNS information similar to a terminal with
the "Dig" command


src-tauri/google.com.au_ips.txt
bradleybogg, line 1: This is not an appropriate location.

package-lock.json
bradleybogg, line 1: Do not include this file.

src/components/DigTool/DigTool.tsx
bradleybogg, line 1: This does not conform to the baselined component available at Airbase NG.
Please use this and any baselining component card in trello for more information.

Thank you

bradleybogg (changes requested):

PR #653, opened 2024-05-03
502 feature request netcat listen and connect

basoreeper3007:
**### Feature Implementations**
-Added Listen mode to make the machine listen for connections on a specified port.
-Added Connect mode to act as the connector that connects to an existing listener.


bradleybogg (changes requested):
image
image


Hi @basoreeper3007 

I have attempted to run the code. As seen in first image there is no error.
Second image there is an error that is not replicated within the standard terminal.

Please investigate.

basoreeper3007:
@bradleybogg I have fixed the syntax bug with the Connect feature.

src/components/Netcat/Netcat.tsx
```diff
@@ -48,16 +51,94 @@ const NetcatTool = () => {
...
+                        // Deactivate loading state
+                        setLoading(false);
+                    });
+
+                break;
+
+            case "Connect": //Connects to an nc listener, nc syntax: nc -v <ip address> <port number>
+                let command = `nc${verboseFlagWithSpaceAndDash} ${values.ipAddress} ${values.portNumber}`;
+
+                //CommandHelper.runCommandGetPidAndOutput("nc", args, handleProcessData, handleProcessTermination)
```
michaeljpigott, line 120: Since you are not using this line, please delete. 

- [ ] Delete commented out line.

src/components/Netcat/Netcat.tsx
```diff
@@ -48,16 +51,94 @@ const NetcatTool = () => {
...
+                CommandHelper.runCommandGetPidAndOutput(
+                    "bash",
```
michaeljpigott, line 122: I am not sure why you need to use bash here. The command nc <IP address> <port> should work without bash. It's probably worth a comment to explain. 

- [ ] Comment required to explain the use of bash.

src/components/Netcat/Netcat.tsx
```diff
@@ -48,16 +51,94 @@ const NetcatTool = () => {
...
+            // Cancel the loading overlay. The process has completed.
+            setLoading(false);
+        },
+        [handleProcessData] // Dependency on the handleProcessData callback
+    );
+
     const onSubmit = async (values: FormValuesType) => {
         //Starts off with the IP address after netcat
         //Ex: nc <ip address>
         let args = [``];
```
michaeljpigott, line 87: Not sure why there are backticks. Have you tried what happens when you remove them and leave it as an empty array?

michaeljpigott (changes requested):
Hi there, 

I have tested Netcat on my local computer and it works. I am not sure exactly why you need to use bash to use the connect functionality, so it would be good to add a comment. 

I am interested to know why there are backticks in the initial setting of the args variable. It does seem a bit strange, but I know that isn't your code. 

I also recommend removing the original CommandHelper line in the connect case now that it is commented out.

basoreeper3007:
@michaeljpigott I removed the comment line that wasn't being used. The connect feature can run without bash, I only used it as it helped me avoid syntax errors (see my comment). I left a comment explaining why I used bash there. The backticks don't need to be in the args[] initialization, so I took them out (that was from the previous student).

michaeljpigott (approved):
Thank you Bas for making those changes. Your commit comments were excellent and easy to follow. 

I approve for this pull request to be merged into main pending approval from the application testers. 

Kind regards 

Michael

PR #634, opened 2024-04-24
407 feature request fcrackzip tools

namanbkshi:
This Pull request is generated with the intentions of solving two issues namely: 
1. 407 feature request fcrackzip tools #634 and 
2. [Bug Fix (Baseline)]: Fcrackzip #586

For the feature request:

1.Set 'Show cracked password' to be enabled by default: Initialized the checkedUnzip state with true, so the checkbox for displaying the cracked password is enabled by default.

2. Ensure Charset List remains active: Removed logic that disabled Charset List options based on other checkbox states. Now, all Charset List options remain active and selectable regardless of other option states, ensuring consistent usability.

3.Adjusted the output behavior to always display the cracked password, especially when using the Dictionary Attack Method.
   - Implemented parsing of the command output to extract the cracked password using a regular expression.
   - If a cracked password is found, it is now displayed in the output along with the rest of the output.
   - Additionally, if the "Show cracked password" option is selected, the cracked password is saved to disk.

Improved the baseline by improving the comments & docstrings along with implementing the following changes:

Certainly! Here's a breakdown of the changes in bullet points:

- **Imports Refactoring**:
  - Updated import statements to reflect changes in dependencies.

- **Component Structure**:
  - Retains the same state variables and form handling logic.

- **Dependency Check**:
  - Introduces a new state variable `dependencyAvailable` to track dependency availability.
  - Asynchronously checks for dependency availability on component mount.

- **Dependency Availability UI**:
  - Displays a message indicating dependency unavailability if `fcrackzip` is not found.

- **UI Improvements**:
  - Loading overlay and cancel button are now rendered using a separate `LoadingOverlayAndCancelButton` component.
  - User guide rendering logic moved to a separate `RenderComponent` component, which receives props for title, description, steps, tutorial link, and source code link.

- **Clear Output**:
  - Retains the function to clear output for users.

PR #661, opened 2024-05-07
bugfix initial commit

gssravanthi31:
This PR resolves the issue with fcrackzip installation error. Added fcrackzip in the dependencies.json.

### DIFF
--- a/install-update-media/dependencies.json
+++ b/install-update-media/dependencies.json
@@ -15,6 +15,7 @@
         "arjun",
         "parsero",
         "sherlock",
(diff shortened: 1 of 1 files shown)


michaeljpigott:
Hi @gssravanthi31 

Unfortunately we no longer add tools to dependencies.json. Rather, we should use the checkAllCommandsAvailability module, which is covered by baselining. fcrackzip baselining work has been implemented by pull request #634 so you may need to cancel this pull request. 

Kind regards

Michael

PR #662, opened 2024-05-07
582-bug-fix-baseline-dnsmap

Cyriacjose:
Using Rendercomponent instead of user guide
Adding inline and other important comments
Aligning Code for better presentation and removing some structure of grammar issues


michaeljpigott:
Hi @Cyriacjose you will need to remove the following files from your pull request before I commence the review:
* install-state.gz
* yarnrc.yml
* yarn.lock. 

A baseline bug fix should only require changes to tsx file for the tool that is being baselined. 

Kind regards

Michael

PR #693, opened 2024-05-15
Cancel Operation implemented for Zerologon attack vector

mariarose-francis:
Implemented and tested Cancel operation and is working as expected.


src/components/ZeroLogon/Zerologon.tsx
JiatengTao, line 1: Tested with adding time sleep to the Python script, the cancel function works.

JiatengTao (approved):
Tested with adding time sleep to the Python script, the cancel function works.

PR #714, opened 2024-08-03
Find Offset attack vector UI changes and fixes

Rewniz:
UI changes based on issue #616 to FindOffset attack vector. UI changes include a description of the attack vector, what the exploit does, libraries and dependencies, improved description and steps, valuable note that only ELF files are valid inputs, fixed the incorrect filepath example to an actual working filepath, fixed grammer and missing full stops, fixed typo, added a note to visit walkthrough tab for a guide on the attack vector and how it can be used (guide will be created as next project).

### DIFF
--- a/src/components/FindOffset/FindOffset.tsx
+++ b/src/components/FindOffset/FindOffset.tsx
@@ -8,15 +8,18 @@ import { SaveOutputToTextFile_v2 } from "../SaveOutputToFile/SaveOutputToTextFil
 
 const title = "Find offset";
 const description_userguide =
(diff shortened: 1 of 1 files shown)


michaeljpigott (changes requested):
Hi @Rewniz 

The changes here do not seem to meet the requirements of the related issue, #616 which is to baseline the UI. 

For an example, check out the WhatWeb tool. Each tool should be using tabs for User Guide, Configuration and Tutorial. 

I have added the baseline checklist document to your Planner card for reference.

PR #719, opened 2024-08-04
- Baseline fixed - 580 - dmitry

Oliver2049:
* Added required imports
* Added the appropriate docstring to the functions and interface
* Changed the state variables to camelCase
* Added inline comments to all react hooks and to all component constant
* Removed the Userguide(title,description_userguide) in Stack as it is no longer required
* Added the steps for user guide in component description
* Add required component state variables
* Added sourceLink and updated it with the appropriate link
* Added empty string in tutorial
* Fixed grammar issues                
* Added the render component  for user guide which was not there in the code


michaeljpigott:
@Oliver2049 you will need to remove install-state.gz and yarn.lock from this pull request. Thank you.

PR #724, opened 2024-08-06
708 feature request baseline arpaname

Lost-Rain:
Completed baseline checklist by adding or revising comments


michaeljpigott (changes requested):
Hi @Lost-Rain 

Just one tiny change, I recommend changing "a" in Arpaname to capital "a" in step 2 to make it consistent with the tool's heading. 

I tested the tool and it works after installing bind9. 

Michael

michaeljpigott (approved):
Thank you @Lost-Rain for making that change.

PR #764, opened 2024-08-19
758-bug-attack-vectors-landing-page-not-sorted-alphabetically

namanbkshi:
I added a line to sort the attackVectors array using the JavaScript sort method, which compares the name properties of the objects.

The localeCompare method ensures the sorting is done in a locale-sensitive manner, which is generally a good practice when dealing with strings.

Mapping Sorted Vectors:

The sorted array is then mapped to ToolItem components as before.

This ensures that the attack vectors are displayed in alphabetical order on the landing page.


michaeljpigott:
Good work @namanbkshi. I have merged into main. Thank you @github4phil for reviewing. 

@namanbkshi do not forget to close the related Microsoft Planner card. 😊

PR #818, opened 2024-08-31
791 feature request unicornscan tool implementation

SethJ1:


src/components/Unicornscan/Unicornscan.tsx
```diff
@@ -0,0 +1,238 @@
...
+        // Include additional flags if they are set
+        if (rate) args.push(`-r ${rate}`);
+        if (sourcePort) args.push(`-e ${sourcePort}`);
+        if (sourceIP) args.push(`-g ${sourceIP}`);
+        if (interfaceName) args.push(`-i ${interfaceName}`);
+        if (ports) args.push(`-p ${ports}`);
+        if (verboseMode) args.push("-v");
+
+        CommandHelper.runCommandWithPkexec("unicornscan", [...args, "-I"], handleProcessData, handleProcessTermination)
+            .then(() => {
```
michaeljpigott, line 133: I found that you need to add the output and pid variables here to show the command in the output block. 
```suggestion
            .then(( { output, pid } ) => {
```

src/components/Unicornscan/Unicornscan.tsx
```diff
@@ -0,0 +1,238 @@
...
+                setLoading(false);
```
michaeljpigott, line 134: I found that the loadingOverlayAndCancel screen was not appearing, so copied Aircrack-NG, which also uses runCommandWithPkexec. I found that once I made these changes, the loading screen started to appear. 
```suggestion
               setOutput(output);
                setAllowSave(true);
                setPid(pid);
```

src/components/RouteWrapper.tsx
```diff
@@ -624,6 +625,14 @@ export const ROUTES: RouteProperties[] = [
         description: "Captures data packets' information and displays in different formats.",
         category: "Network Scanning and Enumeration",
     },
+    {
+        name: "Unicornscan",
+        path: "/tools/Unicornscan",
+        element: <Unicornscan />,
+        description:
+            "Unicornscan is essential for scanning servers and hosts to see what available ports are being utilized for network communications.",
```
michaeljpigott, line 633: Changed to Australian spelling - utilised with an "s". 
```suggestion
            "Unicornscan is essential for scanning servers and hosts to see what available ports are being utilised for network communications.",
```

src/components/RouteWrapper.tsx
```diff
@@ -624,6 +625,14 @@ export const ROUTES: RouteProperties[] = [
...
+        category: "Network Scanning",
```
michaeljpigott, line 634: I think this should be Network Scanning and Enumeration. 
```suggestion
        category: "Network Scanning and Enumeration",
```

michaeljpigott (changes requested):
Hi @SethJ1 

This is looking really good. I just found that the loading overlay was not appearing when I ran the tool, and the actual command itself wasn't appearing in the output block. 

I have made a couple of suggestions, which I think may fix these problems. 

Kind regards

Michael

michaeljpigott (commented):
Hi @SethJ1 

Thank you for making those changes. 

I just noticed that when I enter Metasploitable's IP address into the IP address field and choose TCP Scan, nothing appears to happen:

image

The same thing happens when I enter the exact same command into the Terminal. 

I wonder if it needs some extra parameter for it to run? I noticed on the [Kali page](https://www.kali.org/tools/unicornscan/), it has "a" after  the IP address:

image

Kind regards

Michael

SethJ1:
Sorry @michaeljpigott it seems when I was testing the tool I used Verbose mode every time not realising that without that it wont actually show any of the ports. I have made it so that now it is always in interactive verbose mode. The 'a' flag selects all ports on the IP which is usually default but I shouldn't need it as I have a field to select the port range. Thankyou.

michaeljpigott (approved):
Hi @SethJ1 

Thank you for making that change. I have tested and the tool works well. I approve this pull request. 

Kind regards

Michael

PR #819, opened 2024-08-31
Bug fix  Dirb tool describtion

abasb123:
The description of Dirb has been changed and added a detailed description of the tool.


Oliver2049:
Hi @abasb123,
I believed this bug was related to another file not the description of this file, and I already fixed it with other tools as well.

abasb123:
> Hi @abasb123, I believed this bug was related to another file not the description of this file, and I already fixed it with other tools as well.

Hi @Oliver2049 , Which is the other file you are mentioning.

Issue #692, opened 2024-05-15
[Bug]: InstallationModal continues to appear when the command is different from the name the application


michaeljpigott:
### Name of Bug

InstallationModal continues to appear when the command is different from the name the application

### What Happened?

When using the tool RainbowCrack, the InstallationModal keeps appearing because the name of the program (rainbowcrack) is different from the name of command (rcrack). 

When the CommandAvailability tool uses the which command to look for rainbowcrack, it cannot find it despite rainbowcrack being installed and the rcrack command being available. 

It also affects the arpaname tool. 

### Steps to Reproduce

1. Open the rainbowcrack tool. Install rainbowcrack if required. 
2. Go back to the tools menu. Re-select rainbowcrack. 
3. The Installation Modal will still appear despite rainbowcrack being installed.

CAfitech:
This issue has been fixed #825 pull request hence why I am closing this closing the issue.

PR #825, opened 2024-09-01
Issue #692 Fix

Nicko64:
The dependencies variable for each of the tools is used by the Deakin Detonator Toolkit to verify that a given tool is installed on a system. The dependencies variables for the rainbowcrack and arpaname tools were incorrect, preventing users for accessing the tools regardless of if they were correctly installed or not. This pull request has fixed the values of the variables so that the rainbowcrack and arpaname tools work correctly.


michaeljpigott:
Hi @Nicko64 , do not forget to close the corresponding card in Planner now that this commit has been merged into main. 

Thanks, 

Michael

PR #808, opened 2024-08-30
757 bug no such file or directory when trying to access files saved in usrshareddt

anjojoseph:
Hi Michael,
I have updated the readme file about the script execution.
Kindly review and let me know in case any modification is needed.
Thank you.

PR #831, opened 2024-09-03
801 bug smbghostscanner tool misnamed

connorwjordan:
Renamed related files
Corrected naming to reflect the proper name SMBGhostscanner.

To use must re install exploits via method in pr #808 as it has been renamed.


slpysiren (approved):
Looks good, nice catch!

PR #500, opened 2024-03-26
472 file uploader

bradleybogg:
Pull Request Title:

Feature: Implement File Upload Functionality for DDT

Pull Request Description:

This pull request introduces a crucial feature to DDT, enabling users to upload files directly within the application interface. This functionality becomes necessary due to the restricted access to directories from which the application can read for security.

Changes Made:

Added File Upload Capability: Implemented the necessary frontend components to allow users to upload files.

Backend Integration: Established backend scripts in rust to handle file uploads securely.

File Storage: Implemented a mechanism to store uploaded files in a known location accessible to the application.

Amended rainbowcrack to allow upload of a file.

Uplift mantine to 5.10.5 (latest version 5)

PR #836, opened 2024-09-03
File uploader

slpysiren:
Seeing if creating a new branch and re-adding dependencies fixes the issues in #500


slpysiren:
Passing checks! Hopefully everything looks good now. To get this to work I essentially copied everything over from the other branch but added the uuid dependency with yarn and npm add (I'm not sure why I had to do both or if I actually did have to). I also uninstalled volta and yarn from my VM and reinstalled yarn globally with npm. Hard to say which of these was the fix but it may be worth raising an issue about whether or not volta should be advised in the installation.

JacksonGIThuba (approved):
File Handler
the generateFileName function creates a unique file name based on the inputted/provided file along with the current data and time as well as a UUID. The generateFilePath function generates a file path based on the string the user provides.

File Picker
Adding the file picker handles file selection and uploading within a react application. using the FileInput component to allow users to select files, also allowing multiple files to be selected. it then stores the files using the fileNames and setFileNames variables. The handleChange function processes the selected files converting them to arrays, limiting the number of files that can be uploaded with maxFileNum, then calling saveFile to save each of the selected files. Good error messages for UI occurred errors. 

For Rainbowcrack.tsx
added new imports for the new filepicker file (FilePicker and generateFilePath). Then constructing arguments for the tool based off the users inputs, clearing the tool before adding the arguments each time. Saving the file and creating an upload path.

For yarn.lock
Updated versions of plugins and packages, this included changing the urls, integrity hashes and dependencies. This was done for various packages, along with their dependencies. Also updated sources registry from yarn to npm's registry.

Looks good, good work

michaeljpigott:
Hi @slpysiren, when I run `git fetch origin pull/836/head:<name>` to test the code, it asks me to run `yarn install` which is fine. However, when I run `yarn install` it fails with the following message:

image

Do you have ideas about what I should do?

slpysiren:
At a glance I'm not sure but I do notice your yarn version is 4.x.x which is the one I was having issues with.

If you paste that log it mentioned I can have a look later.

michaeljpigott:
@slpysiren This is a copy of the log:

  GNU nano 8.1                                                             /tmp/xfs-dbbf6a33/build.log                                                                      
# This file contains the result of Yarn building a package (node@npm:22.7.0)
# Script name: preinstall

node:internal/modules/cjs/loader:1148
  throw err;
  ^

Error: Cannot find module '/home/kali/Deakin-Detonator-Toolkit/node_modules/node/bin/node'
    at Module._resolveFilename (node:internal/modules/cjs/loader:1145:15)
    at Module._load (node:internal/modules/cjs/loader:986:27)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:174:12)
    at node:internal/main/run_main_module:28:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v20.16.0

package.json
```diff
@@ -31,17 +31,20 @@
         "@mantine/spotlight": "^5.3.0",
         "@tabler/icons": "^1.95.1",
         "@tauri-apps/api": "^1.0.2",
+        "node": "^22.7.0",
```
BrennanHarrell, line 34: Hi @slpysiren,I encountered the same error as Michael when trying to build the DDT tool (Error: Cannot find module). It seems the issue is due to the addition of the Node version in package.json. Removing this line resolved the problem for me.

slpysiren: Ah good catch, don't remember why I added node as a dependency anyway so removing should fix. Thanks!

michaeljpigott: I can confirm that it is working for me now too 😊

PR #848, opened 2024-09-06
changed the variable names

shreya10-chavan:
I have changed the variable ip1 and ip2 to ipGateway and ipTarget, because ip1 and ip2 were not declared.


basoreeper3007 (approved):
Very good!

shreya10-chavan:
Thank you!

PR #885, opened 2024-09-11
834 Install script does not include install_exploits.sh

slpysiren:
Added a line install the `setup_ddt.sh` script that will also run `install_exploits.sh` so the exploits will be located in their intended location. I'm adding this into the main installation script as many contributors run into the issue of exploits not being located where they should be as they have missed the extra installation step. This will have it automatically run to streamline installation further.

### DIFF
--- a/install-update-media/setup_ddt.sh
+++ b/install-update-media/setup_ddt.sh
@@ -59,5 +59,8 @@ fi
 # Install Yarn dependencies
 yarn install
 
+# Move exploits to DDT directory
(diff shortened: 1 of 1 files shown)


Oliver2049:
Hi @slpysiren 
I tested and it worked perfectly fine.
Regards,
Oliver

Oliver2049 (approved):
Hi @slpysiren
I tested and it worked perfectly fine.
Regards,
Oliver

PR #886, opened 2024-09-11
833 Non-popup search box non-functional

slpysiren:
There were a couple ways this could be fixed but I decided to have the unused search box be unfocused. Originally I was trying to have it unfocused when the search overlay closed but I couldn't get this to work so I opted for it to unfocus after opening the search (Spotlight) overlay.

It is worth noting in newer versions of mantine this component has a prop that disable returning focus.

### DIFF
--- a/src/components/Spotlight/Spotlight.tsx
+++ b/src/components/Spotlight/Spotlight.tsx
@@ -58,7 +58,6 @@ function Search() {
             searchPlaceholder="Search..."
             shortcut="mod + shift + 1"
             nothingFoundMessage="Nothing found..."
-            returnFocus={false}
         >
             <SpotlightControl />
         </SpotlightProvider>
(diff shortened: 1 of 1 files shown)


src/components/Spotlight/Spotlight.tsx
```diff
@@ -4,16 +4,29 @@ import { openSpotlight, SpotlightProvider } from "@mantine/spotlight";
 import { IconSearch } from "@tabler/icons";
 import { useNavigate } from "react-router-dom";
 import { RouteProperties, ROUTES } from "../RouteWrapper";
+import React, { useEffect, useRef } from "react";
```
michaeljpigott, line 7: Looks like you did not end up using these so you can probably remove. 
```suggestion

```

michaeljpigott (changes requested):
Hi @slpysiren 

I have tested the search functionality and your change works well. 

I recommend removing the importation of useEffect and useRef because it looks like you did not end up using them. 

Kind regards

Michael

src/components/Spotlight/Spotlight.tsx
slpysiren: Ah yep, removing.

michaeljpigott (approved):
Hi @slpysiren 

Thank you for making that change. I approve this pull request. 

Kind regards

Michael

PR #889, opened 2024-09-12
Crackmapdelete

JacksonGIThuba:
Removed the outdated and unmaintained tool Crackmapexec


Siddh2001:
I'm pretty sure you're not supposed to either change or commit the yarn file.

michaeljpigott (changes requested):
Hi @JacksonGIThuba 

This is looking good, you just need to remove crackmapexec has a scoped command from tauri.conf.json. 

Kind regards

Michael

GoldenBanksia:
PR request closed as changes unable to be resolved. 

Tested with only the above changes, with possibly opening a separate bug fix for tauri.conf.json file change as requested above, but this breaks dependency checks for the entire application. 

Issue still outstanding as of the time this is closed. Double checked through kali and source documentation. New issue opened #1132

PR #893, opened 2024-09-13
Baselined bed tool

gssravanthi31:
This PR solves the issue. #575


Cyriacjose (changes requested):
Hi @gssravanthi31,

One more change is still needed in this code.

We are using "Const" instead of "let". Your form hook which handles the input needs to be changed from let to Const. Refer to the screenshot below:
GITHUB

Other than this one issue, great work overall.

Regards

Cyriacjose (approved):
Tested on local environment working as intended.

Approving the pull request. Well done @gssravanthi31 .

PR #909, opened 2024-09-14
Fixed inconsistent capitalisation and typo

Lost-Rain:
Fixed a few mistakes in the tool list:
In the "Category" column for "SearchSploit", there's a typo in "Vulnerabilityy" that was changed to "Vulnerability"
In the "Category" column for "Rainbowcrack", there's a typo in "Password cracking and Authentication testing". It should be capitalized as "Password Cracking and Authentication Testing" to match the format of other entries.


Jatkin41 (changes requested):
I believe “Testing” should also be capitalised but otherwise looks good

Lost-Rain:
Good catch, I missed that one. Fixed.

Jatkin41 (approved):
Fixes look good

PR #912, opened 2024-09-15
photon tool

Tillu6:
Photon is a powerful web crawler and OSINT tool designed for efficient data extraction and analysis. It can quickly scan websites, uncover hidden information, and gather valuable insights. Key features include :-

Data extraction: Extracts URLs, parameters, emails, social media accounts, files, secret keys, and more.
Subdomain discovery: Identifies subdomains and DNS-related data.
Customizable scanning: Allows for tailored scanning with options like timeouts, delays, seed URLs, and exclusion patterns.
Plugins: Supports plugins for extended functionality, such as vulnerability scanning or specific data extraction tasks.
 so i think it is very useful in our DDT toolkit.


michaeljpigott (changes requested):
Hi @Tillu6 

I am getting the following error when trying to use the tool: 

image

Can you please give an explanation of why all these files need to get added to the codebase? For example, why do we need the Docker file? 

I have noticed that you can install Photon on Kali and use photon as the command rather than pyhon3, that seems like a more orthodox implementation. 

Kind regards

Michael

michaeljpigott:
Hi @Tillu6 can you please confirm that we can close this pull request and delete this branch now that you have submitted pull request #922?

PR #929, opened 2024-09-16
810 references page

mariarose-francis:
Added all the tools and attack vectors to the references page alphabetically. Inserted appropriate descriptions and documentation links for each tool.


mariarose-francis:
Wrong commit

PR #956, opened 2024-09-21
955 - Bug fixed - add scope for Netcat

Oliver2049:
Add scope for Netcat


michaeljpigott (approved):
Hi @Oliver2049 

This looks good to me. Thank you for picking it up. I suspect the command was re-added as part of the CloudBrute pull request though. 

Kind regards

Michael

PR #978, opened 2024-11-14
Updated PT-GUI installation instructions

jdb2002:
I have updated the readme document, to remove the old installation instructions and add the method that fixes the issues that were being experience


cjboyd1999 (approved):
Changes look good to merge. I've edited some terminology to reflect what we'd see in real-world project repositories.

BrennanHarrell (approved):

Issue #988, opened 2024-11-18
[Feature Request]: Improve References page layout readability 


GoldenBanksia:
### Name of Feature

Improve References page layout readability 

### Priority

Medium

### Categories

UI, UX

### Description of Feature

Improve layout from one long list to accordion sections for Gui Development, Tools and Attack Vectors. Display list/s according to selection. 

Similar to the About page, with the four expandable sections for Tools, Attack Vectors, Walkthroughs and References. 

Test with different window sizes and allow for more than one list expanded at a time.

PR #1006, opened 2024-11-20
Implement accordion layout for References page ( issue #988 ).

Harmit3703:
This PR improves the layout of the References page by introducing an accordion component. The accordion will allow users to view sections (such as GUI Development, Tools, Attack Vectors) in a more structured and accessible way, with the ability to expand and collapse different sections based on user preference.


cjboyd1999 (approved):
Changes look good, no issues when testing in the GUI. Good to merge.

Issue #111, opened 2023-03-18
CVE-2022-26134


PushkarGoel:
### Name of Feature

CVE-2022-26134

### Priority

Medium

### Categories

Sec_test

### Description of Feature

In affected versions of Confluence Server and Data Center, an OGNL injection vulnerability exists that would allow an unauthenticated attacker to execute arbitrary code on a Confluence Server or Data Center instance. The affected versions are from 1.3.0 before 7.4.17, from 7.13.0 before 7.13.7, from 7.14.0 before 7.14.3, from 7.15.0 before 7.15.2, from 7.16.0 before 7.16.4, from 7.17.0 before 7.17.4, and from 7.18.0 before 7.18.1.

PR #1017, opened 2024-11-22
Implementation of CVE-2022-26134

connorwjordan:
Implementation of CVE-2022-26134
This is linked to issue #111 
This implements a poc where users can either set a single url or multiple targets from a text file containing a list of targets, also lets users also set a maximum timeout.


cjboyd1999 (approved):
Changes have been reviewed and look good. Approved for merging into main.

PR #1029, opened 2024-11-25
983 openapi integrationv2

russo-christian:
Adds functionality to send a tool's output to ChatGPT for analysis and explanation. Does so via the 'Ask ChatGPT' button which can be added to tools as required; Traceroute has the button added as an example. Will require an OpenAI API key as per the the tutorial in /documentation/developer_information/chatgpt-api-tutorial.md.


cjboyd1999 (changes requested):
As noted, please reserve an ADR number and update your documentation to this new number. Once the conflict has been resolved, I will review your WIP further.

documentation/adr/adr_register.md
```diff
@@ -7,7 +7,7 @@
 | 0004   |                                                                                               |          |
 | 0005   | RESERVED                                                                                      |          |
 | 0006   | Implementation of TCPdump into PTGUI                                                          | APPROVED |
-| 0007   |                                                                                               |          |
+| 0007   | Introduction of OpenAPI integration                                                           | PROPOSED |
```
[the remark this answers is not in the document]
russo-christian: thanks Cam, all fixed. I'll run the failing test every now and then.

src/components/AskChatGPT/AskChatGPT.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const [loading, setLoading] = useState(false);
+
+    /**
+     * Method to send a prompt and additional data to ChatGPT and return the response.
+     * @param prompt The prompt/question for ChatGPT.
+     * @param data Additional context or data to send to ChatGPT (likely the output from a tool).
+     * @returns The response from ChatGPT as a string.
+     */
+    const sendToChatGPT = async (prompt: string, data: string): Promise<string> => {
+        const apiKey = import.meta.env.VITE_OPENAI_API_KEY;
```
cjboyd1999, line 20: After adding the API key to the `.env` file, attempting to use ChatGPT on the Traceroute tool failed due to the API key not being set.

This line seems to be looking for a variable named `VITE_OPEN_API_KEY` within the `.env` file. I updated my `.env` file to use this variable name instead and had no issues. I suspect this line can be edited to be:

```
const apiKey = import.meta.env.OPEN_API_KEY;
```

cjboyd1999 (changes requested):
See the below comment regarding a variable name that seems to need an update.

src/components/AskChatGPT/AskChatGPT.tsx
russo-christian: I've updated the .env.example and ChatGPT documentation to include the VITE_ prefix. I believe it stops the client seeing the API key in a client-server context (although I'm no expert). While it doesn't make a difference from our development point of view, I figured it is technically more secure?

cjboyd1999 (approved):
Changes look good and it all works when testing locally after updating the variable names. Approved for merging.

PR #1036, opened 2024-11-26
Proposed ADR-0008: Integrating Suricata for real-time network analysis

TanvirRahmanTaiyeb:


cjboyd1999 (commented):
I've reviewed this ADR along with Andrew D. and Brennan H. and we agree that the addition of Suricata to PT-GUI does not fit with the product's use cases at this time.

PT-GUI's focuses are around education of penetration testing methodologies, tactics and tools. Or in other terms, PT-GUI focuses on the red-teaming side of penetration testing. Suricata is a tool that may be implemented within a network to fulfill an IDPS role, which is more akin to blue-teaming operations.

In terms of its use for scanning or reviewing packet captures, PT-GUI already has several integrated tools to achieve these functions from a red-team perspective (see Wireshark, tcpdump, nmap, etc).

Because Suricata has such a large overhead in terms of ongoing maintenance, and it doesn't align with the current product's goals, this ADR has been rejected for now.

cjboyd1999 (approved):
Changes to the ADR's status being rejected look good, good to merge.

Issue #1008, opened 2024-11-20
[Bug]: Pcap file not recognised in Aircrack-ng tool


bshean1:
### Name of Bug

Pcap file not recognised despite correct path and filename

### What Happened?

When using the WPA option in the Aircrack-ng tool, the Pcap file is not recognised, even though the path and filename are correct. This issue prevents the tool from functioning properly as the Pcap file is a required input.

### Steps to Reproduce

1. Navigate to the Tools menu.
2. Select Aircrack NG and then open Configuration.
3. Set WEP or WPA-PSK to WPA.
4. Enter a valid file path and filename for the Pcap field (ensure it is correct and accessible).
5. Enter any arbitrary value for the wordlist, output file, and AP identifier fields (these do not affect the issue).
6. Observe that the Pcap file is not recognised, even though the path and filename are correct.

Issue #1007, opened 2024-11-20
[Bug]: Text entered in one input box automatically fills other input boxes in Aircrack-ng tool


bshean1:
### Name of Bug

Auto-filling of multiple input boxes when entering text

### What Happened?

When using the Aircrack-ng tool in WPA mode, entering text into certain input boxes (specifically the wordlist, Pcap, or output file) causes all three input boxes to automatically fill out. This issue makes it impossible to properly configure the tool, as all three file inputs are required for successful operation.

Additionally, when using the Aircrack-ng tool in WPA mode, entering text into the wordlist or output file boxes causes the other to automatically fill out as well.

### Steps to Reproduce

WPA:
1. Navigate to the Tools menu.
2. Select Aircrack NG and then open Configuration.
3. Set WEP or WPA-PSK to WPA.
4. Enter any text into the wordlist, Pcap, or output file input boxes.
5. Observe that entering text into one input field causes the other two input fields to automatically populate with the same text.

WEP:
1. Navigate to the Tools menu.
2. Select Aircrack NG and then open Configuration.
3. Set WEP or WPA-PSK to WEP.
4. Enter any text into the wordlist or output file input boxes.
5. Observe that entering text into one input field causes the other input field to automatically populate with the same text.

PR #1037, opened 2024-11-27
Fix for issues #1007 and #1008, additional bug fixes for Aircrack-ng 

bshean1:
Due to trimester length and the overlaps from the bugs, multiple changes have been made.

Fixed GitHub issues #1007 and #1008.
Gave WEP mode the ability to specify packet capture file.
Removed wordlist option from WEP mode as it is not needed.
Removed required from arguments that are not required for the tool to run.
Removed unused and unnecessary placeholders.
Set selectedModeOption to default to WEP to make sure WEP specific options are loaded in the first time the tool is opened.
Commented out securityType options, as it is made redundant by Aircrack-ng performing that automatically.
Added code to reset advanced mode back to default when option is deselected.

Rewrote construct arguments section as mode specific options were not working and to improve code readability.
Nested if statements were and are still not working and the reason is unclear.
A temporary fix was made to reset the form every time the mode is switched.
Code for mode specific options was left in for a future fix.

Other notes:
There is an option for customConfig that was present in the code, however it has not been implemented.

### DIFF
--- a/src/components/AircrackNG/AircrackNG.tsx
+++ b/src/components/AircrackNG/AircrackNG.tsx
@@ -35,8 +35,8 @@ const description = "Aircrack-ng is a tool for cracking WEP and WPA/WPA2 passphr
 const steps =
     "How to use Aircrack-ng:\n\n" +
     "Step 1: Type in the path to wordlist(s) filename(s) including the extension .txt. e.g. 'example.txt'.\n" +
-    "For files containing hexadecimal values, you must put a “h:” in front of the file name.\n";
-"Step 2: Select the target network based on the access point's MAC address.\n" +
+    "For files containing hexadecimal values, you must put a “h:” in front of the file name.\n" +
+    "Step 2: Select the target network based on the access point's MAC address.\n" +
     "Step 3: Type in the name of your text file including the extension .txt. e.g. 'example.txt'.\n" +
     "Step 4: Click 'Start " +
     title +
--- a/src/components/AircrackNG/AircrackNG.tsx
+++ b/src/components/AircrackNG/AircrackNG.tsx
@@ -18,12 +18,12 @@ interface FormValuesType {
     BSSID: string;
     ESSID: string;
     keyFile: string;
-    securityType: string;
     characters: string;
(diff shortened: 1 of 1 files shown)


src/components/AircrackNG/AircrackNG.tsx
```diff
@@ -35,8 +35,8 @@ const description = "Aircrack-ng is a tool for cracking WEP and WPA/WPA2 passphr
 const steps =
     "How to use Aircrack-ng:\n\n" +
     "Step 1: Type in the path to wordlist(s) filename(s) including the extension .txt. e.g. 'example.txt'.\n" +
-    "For files containing hexadecimal values, you must put a “h:” in front of the file name.\n" +
-    "Step 2: Select the target network based on the access point's MAC address.\n" +
+    "For files containing hexadecimal values, you must put a “h:” in front of the file name.\n";
```
cjboyd1999, line 38: Looks like the edit made to this line, changing `+` to `;` breaks the displayed steps in the user guide:
image

The semicolon (;) terminates the assignment of the `steps` variable, so the remaining steps are not included in this assignment. 

Can you change this back to a `+` to include the remaining steps?

cjboyd1999: Noting also that on the following line where step 2 is written, you'll need to remove the leading `+` to follow the format of the other sentences.

cjboyd1999 (changes requested):
As noted in my comments below, please edit the code regarding the User Guide's steps to fix the issue with steps 2+ not being displayed.

src/components/AircrackNG/AircrackNG.tsx
bshean1: Yeah, this wasn't something that I edited, running npx prettier made this change and I'm not sure why but I should have checked it after.

bshean1: It should be fixed now.

cjboyd1999 (approved):
Changes look good, no issues when testing locally. Approved for merging.

Issue #1024, opened 2024-11-24
[Feature Request]: Show 'Value (for writing)' input field only when 'Write Metadata' is selected in Exif tool


Harmit3703:
### Name of Feature

Conditional Visibility for "Value (for writing)" Field

### Priority

Medium

### Categories

UI (User Interface) UX (User Experience)

### Description of Feature

Currently, the "Value (for writing)" input field is always visible in the ExifTool UI, regardless of whether the "Write Metadata" checkbox is selected. This request is to modify the UI so that the "Value (for writing)" input field only appears when the user selects the "Write Metadata" checkbox. This will improve the UI by reducing unnecessary clutter and making the form more user-friendly.

PR #1041, opened 2024-11-28
Implement conditional visibility for the 'Value (for writing)' input …

Harmit3703:
Added the functionality to only show the Value(for Writing) , when we click on the Write metadata. Issue #1024


BrennanHarrell (approved):
Looks good to me, implemented changes work as expected and no issues when testing on my end. Approved for merging.

Issue #992, opened 2024-11-18
[Feature Request]: Add switch for IP4 and IP6 address for the Arpaname tool


GoldenBanksia:
### Name of Feature

Add switch for IP4 and IP6 address in Arpaname

### Priority

Low

### Categories

UI, UX

### Description of Feature

Add a switch option for IP4 or IP6 addresses for the arpaname tool, instead of one IP address input. 

Include IP address validation, and test based on the chosen type.

PR #1047, opened 2024-11-29
[Feature Request]: Add IPv4 and IPv6 Switch for Arpaname Tool (#992)

Harmit3703:
This pull request introduces a new feature to the Arpaname Tool that allows users to choose between IPv4 and IPv6 addresses for reverse DNS lookups. The key updates are as follows:

IPv4 and IPv6 Selection: Added a radio button switch to select between IPv4 and IPv6 address types.
IP Address Validation: Implemented validation for both IPv4 and IPv6 addresses using regex patterns to ensure proper input format.
Form Update: Modified the existing form to handle the IP type selection and validation. The form now allows users to specify whether the IP address entered is IPv4 or IPv6.

### DIFF
--- a/src/components/Arpaname/arpaname.tsx
+++ b/src/components/Arpaname/arpaname.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback } from "react";
-import { Button, Stack, TextInput } from "@mantine/core";
+import { Button, Stack, TextInput, Radio } from "@mantine/core";
 import { useForm } from "@mantine/form";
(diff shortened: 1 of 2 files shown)


cjboyd1999 (changes requested):
It looks as though one of your changes have broken the tool from running properly.

 In the current `main` branch, the tool runs as expected for a given IP:
image

In your branch, the tool appears to get stuck on a "process running" screen:
image

Please review and test your implementation to ensure that it works as properly.

Issue #993, opened 2024-11-18
[Bug]: About page images overflow on smaller window size


GoldenBanksia:
### Name of Bug

Image overflow

### What Happened?

On the About page, changing the window size down to small overflows the images beyond the given card display. Suggested fix is a horizontal scroll bar for images instead of image overflow.

### Steps to Reproduce

Go to About page – Click to learn more about any of the four pages – change the window size to not accommodate the original image size

PR #1054, opened 2024-11-30
Update About.tsx

niloy1997:
To add a horizontal scroll bar for images when they overflow their container, I set the container to overflow-x: auto and ensure the image stays within its width constraints. I resized the browser window to simulate smaller screen sizes. Also confirmed the images do not overflow their container.


cjboyd1999 (approved):
Changes look good. Tested locally and they appear to solve the bug reported in #993. Approved for merging.

Issue #1052, opened 2024-11-29
[Bug]: enum4linux fails to run


cjboyd1999:
### Name of Bug

Enum4Linux Fails to Run

### What Happened?

The integrated tool `enum4linux` does not run as expected when compared to the output of a terminal. As shown below, the command fails when run from PT-GUI but succeeds if run from the terminal:

From PT-GUI:
```
$ enum4linux -U  127.0.0.1



enum4linux v0.9.1 (http://labs.portcullis.co.uk/application/enum4linux/)
Copyright (C) 2011 Mark Lowe (mrl@portcullis-security.com)

Simple wrapper around the tools in the samba package to provide similar 
functionality to enum.exe (formerly from www.bindview.com).  Some additional 
features such as RID cycling have also been added for convenience.

Usage: ./enum4linux.pl [options] ip
```

From terminal:
<img width="1089" alt="image" src="https://github.com/user-attachments/assets/5867e4b4-86e0-49b9-84bc-cadf7a49329c">

Additional note: this bug appears to be present in the `main` branch and was found whilst testing changes made to the `enum4linux` tool in https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/pull/1044.

As the issue does not appear to be caused by the PR itself, the PR has been approved for merging and will be merged into `main`.

### Steps to Reproduce

1. Run `enum4linux` from PT-GUI
2. Run `enum4linux` from terminal
3. Compare the different results; terminal works as expected, PT-GUI fails to execute

sreedharashwin:
Yes, this is when there is no input provided for the second text box (parameters), there is no check to see if the value exists. Adding that validation now.

PR #1057, opened 2024-11-30
Fixing Enum4Linux to run on DDT

sreedharashwin:
Fixing bug for #1052

So the problem here is that the arguments are not passed properly to the shell. If we see the below, there is a space between -U and the hostname - 
image

This happens when the input for the parameters textbox, there is no validation present for this and hence creates a space during command creation.

As a fix added a check to see if value exists before being made part of args var.

Works fine now - 

image


BrennanHarrell (approved):
As noted, this fixes the issue noted in #1052. Enum4linux now produces the same output as the standalone tool, so I approve this pull request.

PR #1059, opened 2024-12-01
Implemented the fix for the crash after review

Harmit3703:
This pull request introduces a new feature to the Arpaname Tool that allows users to choose between IPv4 and IPv6 addresses for reverse DNS lookups. The key updates are as follows:

IPv4 and IPv6 Selection: Added a radio button switch to select between IPv4 and IPv6 address types.
IP Address Validation: Implemented validation for both IPv4 and IPv6 addresses using regex patterns to ensure proper input format.
Form Update: Modified the existing form to handle the IP type selection and validation. The form now allows users to specify whether the IP address entered is IPv4 or IPv6.

UPDATE: I have made the fix for the Arpaname IP_switch feature which was causing the tool to crash repeatedly. Requesting you to have a look at this new PR and make the necessary review/approval for the PR.

### DIFF
--- a/src/components/Arpaname/arpaname.tsx
+++ b/src/components/Arpaname/arpaname.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback } from "react";
-import { Button, Stack, TextInput } from "@mantine/core";
+import { Button, Stack, TextInput, Radio } from "@mantine/core";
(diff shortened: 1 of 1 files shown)


cjboyd1999 (approved):
Changes look good. No issues when testing locally. Approved for merging into `main`.

PR #1062, opened 2024-12-02
Updated Mantine to v7.14.3

jdb2002:


GoldenBanksia:
@jdb2002 changes to package.json file will be configured with a Dependabot CI/CD auto-update for dependencies implementation

PR #1063, opened 2024-12-02
Rtgen fixes

bshean1:
- Removed hardcoded parameters.
- Added additional parameters that rtgen requires.
- Full command now appears in the terminal.
- Commented out SaveOutputToTextFile as it is not compatible with this tool.

Additional Notes:
- Rainbow tables are saved to /usr/share/rainbowcrack, there is no built-in way to change this.
- User guide needs to be updated with new parameters
- As a consequence of showing the command in the terminal, 'pkexec' also appears before it. I am unsure how to remove this.


cjboyd1999 (approved):
Changes look great. No issues when testing locally. Approved for merging.

PR #1066, opened 2024-12-02
Feature/url validation dnsrecon

Harmit3703:
This pull request adds URL validation to the DNSRecon tool. It ensures that only correctly formatted URLs are accepted.

Changes Made:
Added a regex-based URL validation for input.
Error message for invalid URL formats.
Reason for Change:
Prevents invalid URLs from being submitted, improving the tool’s reliability.

### DIFF
--- a/src/components/Arpaname/arpaname.tsx
+++ b/src/components/Arpaname/arpaname.tsx
@@ -1,5 +1,5 @@
 import { useState, useEffect, useCallback } from "react";
-import { Button, Stack, TextInput } from "@mantine/core";
+import { Button, Stack, TextInput, Radio } from "@mantine/core";
 import { useForm } from "@mantine/form";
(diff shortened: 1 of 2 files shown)


cjboyd1999 (changes requested):
Two issues found in these changes:

1. `arpaname.tsx` - the tool does not run as expected when compared to `main`. The process freezes on a "process running" screen that cannot be escaped.
2. `Dnsrecon.tsx` - your regex looks good for URL's, but the tool does not take a URL, just a domain. For example, you're asking for a URL such as `https://google.com`. If you try to run this in the tool, it throws an error. It's only looking for a domain, e.g. `google.com`. Please update the regex along with the user guide and configuration sections to reflect that its looking for a domain, not a URL.

Harmit3703:
Hey @cjboyd1999 , can you check my new commit with updated changes for the dnsrecon tool and i dont know how the arpaname tool code came with this pull request so please ignore that.

PR #1073, opened 2024-12-03
Fix for trivy workflow cache

BrennanHarrell:
The current implementation of the Trivy workflow was not effectively utilizing the existing cache during the security check. This update addresses that issue:

- The security check now correctly attempts to restore the Trivy cache created by the update-trivy.yml workflow. This workflow downloads the required Trivy DBs daily at midnight and generates a cache that is used during the security check.
- If for whatever reason the cache is not found, the security check workflow will download and extract the required compressed DB files to prevent the workflow from failing.
- Downloading the compressed DBs also seems to fix the "TOOMANYREQUESTS" issue previously seen. Meaning this new implementation will also not experience this issue even if the cache is not found.


cjboyd1999 (approved):
Changes approved for merging. Hopefully this fixes the common `TOOMANYREQUESTS` error.

PR #1078, opened 2024-12-04
The Harvester tool fix

connorwjordan:
The Harvester tool was not working due to the commands being removed from the tauri.conf.json file, the command has now been re-added.
Also added an if for checkAdvanced for the advanced option arguments in the onSubmit, and removed unneeded type prop.


cjboyd1999 (approved):
Changes look good. No issues when testing locally. Approved for merging into `main`.

PR #1081, opened 2024-12-04
Mantine update

jdb2002:
Updated project to latest release of mantine. Attempting to pass CI checks


jdb2002:
> Hey Jordan,
> 
> I’ve been reviewing the relevant Mantine documentation, and I wanted to double-check with you—are you sure that @mantine/prism still works with Mantine v7? I highly recommend reviewing the changes introduced in Mantine version 7, particularly how these changes affect prism and createStyles. This is likely related to why #1062 was failing the test and build CI checks.

Hey Brennan! Yeah I was reading up on that yesterday. It’s strange because when I updated to version 7 on my local build, when yarn install was run, it would output about an error with mantine/prism, but the gui would still launch and work as expected. I’m in the process of writing up about why we can’t update mantine for the handover documentation

GoldenBanksia:
Same closing comment as #1062, we do need to update package.json file, but this will be done by Dependabot as required

PR #1082, opened 2024-12-04
Filepicker netcat implementation

BrennanHarrell:
This pull request integrates the FilePicker component into Netcat, specifically enhancing the 'Send File' functionality by allowing users to easily locate and select the desired file.  It should however be noted that this will only work if the file is located in /home/kali. The component has not been implemented for the 'Receive File' function, as it would cause that functionality to break.


GoldenBanksia (approved):
Tested on local machine for each tool, functionality works as expected for four updated tools Hydra, JohnTheRipper, Netcat and RTsort. 

No issues with dependencies, introduced errors or breaking tools. 

Other issues with tools not related to this PR, this is compatible with current version. To be merged + closed.

PR #1128, opened 2025-03-05
Create improvement_request.yaml template for new issues. 

GoldenBanksia:
Add new issue template in for improvements, distinct from feature requests.


TanvirRahmanTaiyeb (changes requested):
Just a few small things I caught before this can be merged.

First, there’s a typo in the description field—"ugrade" should be "upgrade." Just a quick fix, but good to clean that up.

Also, in the categories placeholder, "Accessability" is misspelled—it should be "Accessibility."

For the priority field, the description could be clearer. Instead of "What is the priority of this feature?" something like "Select the priority level for this improvement request" would make it more direct.

TanvirRahmanTaiyeb (approved):
Looks great! Merged into main.

PR #1144, opened 2025-03-23
Application welcome / disclaimer acknowledgment page​

Nuds94:
-Added pop up for user to click and accept before continuing
-Added blur to background behind box
-Added logo and message to box and aligned

Note: need to look into stopping user from just clicking off the box to close the message


Nuds94:
upload linked to feature request - Application welcome / disclaimer acknowledgment page #1145

GoldenBanksia:
Looks good 
- the implementation is simple + effective, and in keeping with the existing standards for rendering and templating. 
- Nice code comments 

Feedback
- only thing to note is the phrasing should be more specific
"DDT does not condone unethical hacking" - good
"Misuse can lead to jail time,
Hacking is BAD! Don't Hack!"
- Effectively gets the main point across which is what is currently missing. However, for legal purposes, best to be more specific.  Could be something like... 

"Hacking is a crime. This application is for **educational purposes only**. 
Misuse of this application can lead to violation of Australian and/or International Law. By using this application, you agree that when performing a penetration test with this software, you have the authorization of all parties involved. You are solely responsible for managing this authorization. 
Different states and territories within Australia have different laws, always check you are in compliance with your specific location before using this application."  

**Out of the above, the most important bit is 'educational purposes only' and that the 'user agrees they have permission/authorization when using the tools' in toolkit.**  Please add at least these two bits in to your phrasing and I will get this merged for you asap.

Future notes 
- The pop-up relaunches each time a user goes back to the About page, but we would need a backend server to store the 'true/false' value for this, so not a problem at all for now as can't be helped
- We want to add in a full user agreement / T&Cs later but will need a Deakin staff member to help with this

src/pages/About.tsx
```diff
@@ -20,6 +22,35 @@ const AboutPage = () => {
...
+            >
+                <Image
+                    src="src/logo/logo-dark.png"
+                    alt="logo"
+                    width={300}
+                    style={{ display: "block", margin: "0 auto" }}
+                />
+                <Text align={"center"}>DDT does not condone unethical hacking,</Text>
+                <Text align={"center"}> Misuse can lead to jail time.</Text>
+                <Text align={"center"}>Hacking is BAD! Don't Hack!</Text>
```
GoldenBanksia, line 48: Please see comment for review

GoldenBanksia (changes requested):

src/pages/About.tsx
Nuds94: Yea was more there as a place holder and testing, Do we have a confirmed message we want to put in? can update it

GoldenBanksia: No problem, there is an example in the review comment above you can use. Thanks Andrew

Nuds94:
Updated the message and have started looking into the issue where it reopens every time you return to the page

open

GoldenBanksia (approved):
Looks great - changes requested have been implemented. Known issue of pop-up reopening is okay for now, can be solved in a future PR. 

Tested and runs well on local machine. No introduced errors with bugs/dependencies/libraries. Approved to merge into main

PR #1146, opened 2025-03-25
ADR 0014: Remove Dirb from PT-GUI due to redundancy and outdated status

TanvirRahmanTaiyeb:


rng4302 (approved):
The reason given to remove the tool seems logical. So, merging the branch into main.

PR #1149, opened 2025-03-26
refactor: removed Dirb tool routes and references from UI

TanvirRahmanTaiyeb:


GoldenBanksia (approved):
Looks great, tested and runs fine on local system. 
- no issue with updated dependency list (application recognizing when they are installed/not installed for tools as required)
- other tools function fine after tool removal 
- routewrapper file updated correctly
- other functionality used in dirb not removed is used by, or from other tools
- tauri config file all good, did not contain dirb
- references updated + tool removed from list
- current & approved ADR0014 outlines reasoning for this removal

PR #1153, opened 2025-03-31
Updated 'arpaname' and 'arjun' tools to throw an error when dependencies are not fulfilled.

nopenotdaniel:


nopenotdaniel:
Had some issues with prettier updating the yarn lock file. Will fix this on a new branch instead.

nopenotdaniel:
Issues with software versions of mine vs project versions, I'll do a clean install of the VM and just slap the new code for arpaname/arjuntool in there when it's done.

Issue #1120, opened 2025-01-10
[Bug]: Missing Target Input in 'Scan a Range of Ports' Configuration for Hping3


MrunalDas:
### Name of Bug

Missing Target Input in 'Scan a Range of Ports' Configuration for Hping3

### What Happened?

The "Scan a Range of Ports" configuration for the Hping3 tool in the Deakin Detonator Toolkit (DDT) does not have an input field to specify a target IP address or hostname. This makes the feature unusable as the tool cannot resolve an empty target.

### Steps to Reproduce

1.Open the Deakin Detonator Toolkit.
2.Navigate to the Hping3 tool.
3.Select the "Scan a Range of Ports" option.
4.Attempt to start the scan without any option to specify a target.
5.Observe the error in the output: Unable to resolve ''.

**Expected Behavior:**
The "Scan a Range of Ports" configuration should include a field to input a target IP address or hostname, allowing users to specify where the scan should be directed.

**Actual Behavior:**
There is no input field to specify a target IP or hostname in the "Scan a Range of Ports" configuration. As a result, the scan fails with the error: Unable to resolve ''.

**Screenshots:**
![Screenshot 2025-01-10 175420](https://github.com/user-attachments/assets/f86dc2ee-9e8a-490e-bd20-00757483d455)

PR #1194, opened 2025-04-10
Hping3 fix issue #1120

rfrofia:
Updated Hping3 so that when opened it will correctly show input fields as image below: 
hping3

updated "const [selectedOption, setSelectedOption] = useState("");" to "const [selectedOption, setSelectedOption] = useState("Scan a range of ports");"
hping3 change


TanvirRahmanTaiyeb (approved):
Looks good to me. Setting the default value for selectedOption is the right approach here to make sure the input fields are visible when the tool is first opened.

Tested locally and it matches the expected behaviour shown in the screenshot.

Everything looks good from my side. Approving this.

PR #1196, opened 2025-04-10
Setup_ddt.sh update: Check Debian security and repository already added in sources.list 

GoldenBanksia:
Updates setup_ddt.sh file in GitHub repo 
- Currently debian security & repository are added to /etc/apt/sources.list every time curl command or ./setup_ddt.sh is run
- This updates setup file to check if the source is already added 
- Does not add again if this is true (separate check for each source)
- Avoids possible issue with matching #deb http://... file with ^ character
- Prints validation to console for user

### DIFF
--- a/install-update-media/setup_ddt.sh
+++ b/install-update-media/setup_ddt.sh
@@ -2,8 +2,10 @@
 
 set -e
 
-# Add Debian security updates and main repository to sources.list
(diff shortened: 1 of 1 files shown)


TanvirRahmanTaiyeb (approved):
Looks good, this is a clean improvement. The grep check before adding the repos makes the script much safer and avoids any duplicate entries while still giving clear feedback to the user.

Tested locally and everything works as expected. Approving for merge.

PR #1201, opened 2025-04-11
ADR request for implementation of News Feed for cyber awareness

Sanchit-Gt:
This ADR documents the decision to integrate a cybersecurity news feed into the DDT platform. The feature aims to enhance user engagement by displaying real-time threat intelligence and news updates from trusted cybersecurity sources.


TanvirRahmanTaiyeb (changes requested):
The style check for this PR has failed. You need to run the following command to fix the formatting issues and commit the changes to pass the check: "yarn prettier --write ."

For the adr-register.md section please fix the wording. It should be W-I-P meaning Work In Progress, not Under Process. Make sure to follow the same structure and terminology throughout the ADR for consistency.

Also, your ADR-0004.md file is currently empty. This is where you need to write your actual ADR content.

TanvirRahmanTaiyeb (changes requested):
Thanks for writing the ADR, the idea looks good and relevant for PT-GUI.

Few things to fix based on our ADR structure. You can refer to ADR-003 or ADR-0015 for the correct format.

Please replace DDT with PT-GUI everywhere. Avoid using words like "I decided" or "I identified" as ADRs should be written in a neutral way like "This ADR proposes" or "PT-GUI will integrate".

At the top you are missing the standard sections we use in every ADR. Please add Title, Proposal Status and Implementation Status.

In the context bullet point is not necessary. You can write it in a paragraph format like other ADR examples.

Mitigation heading is not really needed here unless there is a critical risk. You can remove it or if you want you can mention it under Consequences section to justify it properly along with Positive and Negative Consequences.

For the Consequences section please follow ADR-0015 style. First write Consequences as the main heading. Then under that use Positive Consequences and Negative Consequences as subheadings.

Also we do not use emojis in any ADR headings. For Pros and Cons section just keep it as Positive Consequences and Negative Consequences.

And for adr-register.md do not remove everything and rewrite it. Just add your line for your specific ADR number only. Make sure not to touch or mess up the other existing ADR numbers and info.

If you have copied or referred to any content from online sources or GPT for this ADR please make sure to change it properly according to PT-GUI guideline and structure. Do not copy paste directly without following the standard.

Once these are fixed it should be good for review. Only when you think it is ready then commit and push the code. After that let the leaders know so it can be reviewed and merged.

TanvirRahmanTaiyeb (changes requested):
You have removed another person's ADR entry from the register. Please do not remove existing approved entries. Also, for the new line you added, it should be written as W-I-P not W-I-P git to match the existing format. The rest of the structure looks fine. Once you fix these, let me know and I will review it again.

TanvirRahmanTaiyeb (approved):

PR #1211, opened 2025-04-14
fixed bug 1156 Setup_ddt.sh duplicate repo

ataklaSec:
-fixed issue of setup_ddt bash file duplicating repositories on every execution.
-ignores commented lines with "#".
-tested and worked.


TanvirRahmanTaiyeb (approved):
Looks good, the issue is fixed properly and the sources.list handling is done in the right way. Now it will not duplicate the lines even if the script runs multiple times. Everything looks fine and tested properly.

PR #1212, opened 2025-04-15
 ADR for Anonsurf 

hades7200:
Created ADR for Anonsurf Functionality


TanvirRahmanTaiyeb (changes requested):
Just went through the PR. Please fix the merge conflict in adr_register.md. The ADR work should be kept separate from feature development like the HomePage changes. For this PR, only adr_register.md and the new ADR file need to be modified. The changes in App.tsx and other files are not necessary and should be removed. Keeping the ADR work focused will make it easier to review. Let me know once it is cleaned up.

PR #1216, opened 2025-04-17
Update ADR-0009.md

TanvirRahmanTaiyeb:
Updated the implementation status as it is already completed.


GoldenBanksia (changes requested):
Looks good - also need to update adr_register to implemented

GoldenBanksia (approved):
Changes implemented in PR #1218 - approved to merge into main

PR #1220, opened 2025-04-18
Fix LoadingOverLay baseline bug for Netcat

Oliver2049:
- Review the baseline to investigate why Loading overlay does not work for Netcat and fix the bug.
- Simplify the baseline code as needed to reduce inconsistency and redundancy.

### DIFF
--- a/src/components/Netcat/Netcat.tsx
+++ b/src/components/Netcat/Netcat.tsx
@@ -146,144 +146,56 @@ const NetcatTool = () => {
      * @returns {Promise<void>} - A promise that resolves when the command execution is complete.
      */
     const onSubmit = async (values: FormValuesType) => {
-        //Starts off with the IP address after netcat
-        //Ex: nc <ip address>
-        let args = [];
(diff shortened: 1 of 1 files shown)


TanvirRahmanTaiyeb (changes requested):
The loading overlay fix looks good and the onSubmit logic is much cleaner now. Just one thing to flag. In the file transfer sections, the use of < and > inside the args array will not work as intended. These are shell operators and not valid arguments, so unless the command is run through something like bash -c, redirection will not happen and the command will fail at runtime.

To keep file transfer reliable, it is better to revert that part back to using a full command string through bash -c like it was originally. Everything else looks good.

Oliver2049:
@TanvirRahmanTaiyeb  Hi I already fix it according to your feedback. 
Regards,
Oliver

TanvirRahmanTaiyeb (approved):
Thanks for addressing the feedback. The updates now look clean and everything is working as expected after testing. Approving this for merge.

PR #1227, opened 2025-04-20
Update ci.yml workflow version and file

GoldenBanksia:
added in new file in case it breaks/doesn't work
- update timeout for all jobs to 30mins instead of default 360
- remove reference to actions/cache as covered by actions/setup-nodev4
- update actions/checkout and actions/setup-node from v3 to v4
- added clearer code comments for future devs 

Dependency job
- update matrix version to 20.x
- update path to dependency cache - was not valid before
- remove install yarn as this is already pre-installed in latest ubuntu github hosted image runner
- remove reference to keys -> as we use yarn package manager (should actually remove package.lock json file asap)

CacheNotFound actionsworkflow
two lock files actionsworkflow

Style job
- updated names for steps for maintainability

Build job
- moved order of job
- added TODO item for future testing

Test job
- added pre-req that build job needs to be passing
- updated names for steps for maintainability

Security job
- except for changing timeout, no changes


TanvirRahmanTaiyeb (approved):
Looks good. The workflow is structured clearly and aligns with the current CI setup. Updates to node versioning, caching strategy, and job dependencies are implemented properly. Approving this.

PR #1229, opened 2025-04-21
ADR-0019: Replace Wifite with Wifite2

kani2003:
- Created ADR for Wifite2 Upgrade (ADR-0019.md)
- Updated the ADR registar.md with the ADR number


TanvirRahmanTaiyeb (approved):
ADR looks fine. Wifite2 is a better fit and the reasoning is clear.

PR #1234, opened 2025-04-23
change package manager to yarn to match project

GoldenBanksia:
-change package manager to yarn, as npm will update package-lock when we want to update yarn-lock file
-reduce PRs to 5

- Realized this isn't actually running currently. Will enable later after review


GoldenBanksia:
Nevermind! npm all good for now - may look into cargo package manager but not important

PR #1244, opened 2025-04-29
Bug 1180

jdb2002:
This is the completed bug fix for the Pkexec cancel button. I have fixed the issue so that the button actually ends the process of the tool when clicked and exits with the correct information in the shell. There are a list of tools that use the Pkexec Cancel button that are entirely implementing the button incorrectly. I will add these to the git issues page and the MS teams board as their own bug fixes. These tools are arp fingerprint, arpscan, urlsnarf, arp spoof, exif and rtgen.


TanvirRahmanTaiyeb (approved):
Tested the cancel button on a few tools and the signal is now handled correctly. The termination message appears as expected and the implementation aligns well with how pkexec-based tools are structured. All looks good to merge.

PR #1245, opened 2025-04-29
Added AI Training Scenario

TanvirRahmanTaiyeb:
- Added AI Training Scenario page and connected it to the sidebar and routes.

- Changed the sidebar label and icon to match the training theme.

- Updated tauri config to allow saving .txt reports from scenario page


jishnu5555 (commented):
Adding the AI Training Scenario is honestly a brilliant upgrade to the PT-GUI project. This feature gives users a hands-on way to understand what kind of scenarios they might encounter with the CVE they've picked, which makes the whole experience way more immersive and practical. It doesn't just generate a scenario, it also throws in useful hints to help users fine-tune their vulnerability reports, which is such a key skill for anyone getting into cybersecurity. I tested this locally and can confirm that the feature works as intended and doesn’t break or interfere with any existing functionality in the toolkit.

jishnu5555 (approved):
Adding the AI Training Scenario is honestly a brilliant upgrade to the PT-GUI project. This feature gives users a hands-on way to understand what kind of scenarios they might encounter with the CVE they've picked, which makes the whole experience way more immersive and practical. It doesn't just generate a scenario, it also throws in useful hints to help users fine-tune their vulnerability reports, which is such a key skill for anyone getting into cybersecurity. I tested this locally and can confirm that the feature works as intended and doesn’t break or interfere with any existing functionality in the toolkit.

Issue #1247, opened 2025-04-30
[Bug]: Loading Screen in Arp Spoof tool


jdb2002:
### Name of Bug

Loading screen not rendered after initating Arp Spoof tool

### What Happened?

Loading screen is not rendering when Start button is clicked leading to the inability to cancel the process. Multiple clicks of the start button leads to opening more instances of the tool, leading to incredibly high cpu usage.

Fix loading screen implementation within the tool's code, so that properly renders and prevents multiple instances running. Change signal value to '2' to match with Pkexec cancel button (bug fix 1180) so that the pop up also allows cancelling the process and outputs the correct information. 

### Steps to Reproduce

Fill fields required to run Arp Spoof then click 'Start'.

PR #1252, opened 2025-05-01
Bug #1247 Loading Screen in Arp Spoof tool

jdb2002:
Fixed code for Arp Spoof tool, and added extra functionality to the pkexec cancel button allowing it to accept a second PID to kill. This argument has been identified in the code for the cancel button and has been assigned a default value. When the Pkexec cancel button is called, a warning does appear if the second PID isn't defined, however the cancel button still works as expected with this warning. I have gone through and added an empty argument to the tools that do call this function to avoid any warnings. The button's function remains unchanged for tools that create a single PID and now also works for tasks that create 2. I have found even more bugs within the tools; aircrack ng, gobuster, netcat, photon and sql ninja so I will add these as their own git issues and fix them in separate pull requests

### DIFF
--- a/src/components/AirbaseNG/AirbaseNG.tsx
+++ b/src/components/AirbaseNG/AirbaseNG.tsx
@@ -210,7 +210,7 @@ const AirbaseNG = () => {
             )}
             <form onSubmit={form.onSubmit(onSubmit)}>
                 <Stack>
-                    {LoadingOverlayAndCancelButtonPkexec(loading, pid, handleProcessData, handleProcessTermination)}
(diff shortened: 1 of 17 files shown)


Sir-Floof (approved):
Request addresses issues with the ArpSpoof tool. Tool's functionality has been improved when comparing before and after the changes. Cancel button is also working on the loading overlay. Loading overlay improvements have been applied to a variety of other tools. Several of these have been tested and behaviour is consistent between them. 

Code comments have also been included which will make it easier for others to understand the changes made. Approved for merging into main.

PR #1253, opened 2025-05-02
CVE-2023-22527

rfrofia:
created .py exploit file along side .tsx gui file and updated RouteWrapper.tsx file to add CVE-2023-22527 to program
wrapper
pyfile
tsxfile


rfrofia:
re-doing after running prettier

PR #1258, opened 2025-05-03
Refactor aircrackng cleanup

bshean1:
- Removed all unused code
- Added function that combined the reset all advanced mode options for code readability and scalability.
- Additional formatting changes

- When testing the changes, it was found that using KoreK mode on non .ivs files, and inputting a bssid with .ivs files without KoreK enabled causes the program to hang. This issue occurs even in the normal Linux terminal. An information box has been added to warn the user about this issue.


TanvirRahmanTaiyeb (approved):
Everything looks good from my side. I tested the functionality thoroughly and the resetModes behavior is working exactly as expected. The korek warning implementation is also clear and helpful. No issues found during testing. Approving and merging this into main now.

PR #1267, opened 2025-05-05
Bug 1256 fix

jdb2002:
The Netcat tool could not be cancelled when it was run. After investigating, I found that the code did not correctly implement the run function, and was not assigning the Process ID to the pid variable. After fixing this, the cancel button works as it now knows what PID needs to be killed when it is clicked


TanvirRahmanTaiyeb (approved):
Tested it locally and everything is working as expected now. The cancel button correctly terminates the Netcat process and the PID assignment is handled properly. Good fix and the code changes are clean. All good from my side.

PR #1272, opened 2025-05-06
ADR-0022: Add feature to temporarily block access to broken tools

bshean1:
- Created an ADR proposal to add a feature to block access to broken or in-development tools.
- Updated the ADR register accordingly.

### DIFF
--- a/documentation/adr/ADR-0022.md
+++ b/documentation/adr/ADR-0022.md
@@ -1,0 +1,36 @@
+## Title
+
+Add feature to temporarily block access to broken tools
+
+## Proposal Status
+
+-   Proposed
+
+## Implementation status
+
+-   Work in progress
+
+## Context
+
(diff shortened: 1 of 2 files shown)


TanvirRahmanTaiyeb (commented):
Considering the current goals of the project and the stage we’re at, this isn’t the right time to introduce this feature. While it makes sense from a usability point of view, it adds unnecessary complexity to our current workflow. Right now, we're already managing broken or unstable tools manually without any major issues. Introducing this would mean extra steps like marking tools as blocked, pushing a PR, waiting for approval, and then requiring everyone to pull the latest changes just to reflect that status all of which feels a bit premature. It also doesn’t align with the current focus of the project, so this ADR has been rejected for now.

jishnu5555 (approved):

PR #1277, opened 2025-05-07
Bug 1273 fix

jdb2002:
The cancel button now renders, as the original code for this tool set the loading state to false when the tool was run, meaning this overwrote the true command found earlier in onSubmit. When I fixed this, the cancel button rendered but was not working. I discovered that during the run command, the Process ID was never passed to the pid variable which is used for the cancel button. This meant the cancel button wasn't able kill the PID as it was passed a null variable. When I fixed this, the cancel button was still not working as expected, and I discovered that this tool was calling the normal cancel button rather than the Pkexec cancel button. Now that it is calling the correct cancel button, I updated the values in the handle process termination section to ensure the correct output in the shell. The cancel button and this tool as a whole are now working as expected.


TanvirRahmanTaiyeb (approved):
Tested locally and everything is working properly. The cancel button now appears and functions as expected. Output handling also looks correct.

PR #1281, opened 2025-05-07
Improveing UI for Netcat

Oliver2049:
- Netcat now will have a better UI with three steps to run a command, which is easier for beginer to follow.


TanvirRahmanTaiyeb (approved):
Looks good. The updated UI is much more beginner-friendly and the flow is easier to follow.
No issues found during testing. Approving and merging this into main now.

PR #1284, opened 2025-05-08
Revert "Bug 1274 fix"

TanvirRahmanTaiyeb:
Reverts Hardhat-Enterprises/Deakin-Detonator-Toolkit#1283

### DIFF
--- a/src/components/Urlsnarf/Urlsnarf.tsx
+++ b/src/components/Urlsnarf/Urlsnarf.tsx
@@ -146,6 +146,7 @@ const Urlsnarf = () => {
         args.push(`-v`, `${values.versusMode}`);
 
         if (selectedListenerInput === "Interface") {
+            setLoading(false); // TODO - Have loading state only be true while inputting password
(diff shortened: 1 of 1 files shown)


TanvirRahmanTaiyeb:
Hi @jdb2002   I’ve reopened the pull request because the issue still isn’t fully resolved. The cancel button is still not working as expected. While everything works fine when accessed through the web version, but when the toolkit opens automatically and the user cancels, the cancel button does not behave as expected.

jdb2002:
Hi @TanvirRahmanTaiyeb, I believe this is not a bug with the cancel button. I have noticed in the code that the placeholder for dropdown box is not working, meaning that it appears one of the two options have been selected, when neither of them have been. When a user clicks the start button it should pass an error that the user hasn't selected a mode and not initiate the loading state. Since the tool isn't actually starting, there is no PID to pass to the cancel button, therefore the loading state doesn't go away. I will look into how to fix the dropdown box

TanvirRahmanTaiyeb:
This issue has been fixed on another PR.

PR #1285, opened 2025-05-08
 Bug fix 1275

jdb2002:
ArpScan wasn't able to be cancelled as the initial code never set the value of the PID, meaning the cancel button wasn't able to kill the process. Infact, this tool never defined a PID variable to be set as the tool's PID.  the variables 'loading' and 'pid' are meant to be passed to the cancel button, however this tool passed "loading, " "' meaning that it was passing a null value to the function. To rectify this, I declared a PID variable, as well as set its value in the run command and passed this variable to the cancel button. On testing, the cancel button was not working as expected so I changed this tool to employ the pkexec cancel button and it is now working as expected.

### DIFF
--- a/src/components/ArpScan/ArpScan.tsx
+++ b/src/components/ArpScan/ArpScan.tsx
@@ -134,7 +134,7 @@ function ARPScan() {
         const args = [`--localnet`, `-I`, values.interface];
 
         CommandHelper;
-        // Execute the Masscan command via helper method and handle its output or potential errors
+        // Execute the ArpScan command via helper method and handle its output or potential errors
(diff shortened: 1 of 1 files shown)


Sir-Floof (approved):
These changes fix the bug referenced and has been confirmed by testing the application before and after. The ARPScan tool is now able to be cancelled with an appropriate message displayed to the user. All other application behaviour remains normal. Approved for merging into main.

Sir-Floof (approved):
These changes fix the bug referenced and has been confirmed by testing the application before and after. The ARPScan tool is now able to be cancelled with an appropriate message displayed to the user. All other application behaviour remains normal. Approved for merging into main.

PR #1287, opened 2025-05-08
Further bug fixing for UrlSnarf

jdb2002:
This fix addresses concerns raised about the ability to launch the tool before selecting options, leading to the loading state initiating and not being able to cancel it. I have fixed the drop down list so that the user must select and option, and then fill the required fields before the tool will start. This means the cancel button works to cancel the tool in all instances


TanvirRahmanTaiyeb (approved):
Just tested this locally and everything worked well. The tool only launches once the user selects an option and fills in the required fields, and the cancel button behaves correctly in all cases now. This fix properly addresses the issue mentioned earlier. You can go ahead and close the GitHub.

PR #1288, opened 2025-05-09
ADR 0024 for Bettercap Functionality

nopenotdaniel:


TanvirRahmanTaiyeb (changes requested):
This ADR isn’t following the proper structure we’ve used in the other ADRs. Please refer to the format and align it accordingly. Also, in the ADR register, avoid writing vague descriptions like “Created an ADR for BetterCap implementation.” Instead, write something more direct like “Integration of BetterCap into PT-GUI.” Once the structure is corrected and everything is aligned properly, feel free to send it in for review again and let me know.

TanvirRahmanTaiyeb (approved):

Issue #1217, opened 2025-04-17
[Bug]: Traceroute Failing to accept multiple additional flags.


MetricDunn:
### Name of Bug

Failure to scan with more than one flag appended to the Traceroute shell.

### What Happened?

Traceroute locks up or returns a a failure message when appended more than one flag to a custom scan.

### Steps to Reproduce

Simply add more than one custom flag to a custom scan.

MetricDunn:
Still working on this. Please do not usurp this task.

jishnu5555:
The related PR has been submitted and successfully addresses this issue. As a result, this issue is now being closed. Thanks for your contribution!

jishnu5555:
The related PR has been submitted and successfully addresses this issue. As a result, this issue is now being closed. Thanks for your contribution!

PR #1292, opened 2025-05-09
Fix for Traceroute issue 1217

MetricDunn:
The alterations I have made to Traceroute.tsx rectify issue 1217.

Custom scans now function as intended, previously they had not appropriately appended the flags used in a custom search and in some cases not at all. Now custom commands function as intended.

Once this error was corrected the custom scan failed to operate with multiple flags. I have also corrected this issue as well. As long as the combination of flags is valid for a traditional Traceroute command then they will work as intended.

I have attached a video demonstrating that the tool is now working as intended.
https://youtu.be/J3IFZgntwx0

### DIFF
--- a/src/components/Traceroute/Traceroute.tsx
+++ b/src/components/Traceroute/Traceroute.tsx
@@ -133,9 +133,11 @@ const Traceroute = () => {
      * @param {FormValuesType} values - Contains the type of traceroute scan (ICMP, TCP, UDP, custom),
      * the destination hostname, and optional custom traceroute options.
      */
+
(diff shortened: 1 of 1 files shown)


jishnu5555 (approved):
Thanks for fixing issue #1217 and for including the demonstration video, it clearly explains the changes made to the "Traceroute" tool. The tool is working as intended, and everything looks good. Great work!
Just a suggestion for future PRs: please link the related issue directly in the PR description to make it easier to trace back.
Overall, this is a solid submission, thank you for your contribution!

PR #1296, opened 2025-05-09
Fix for issue 1259 Nmap tool missing start button

MetricDunn:
I have added a "Run Nmap" button to the first stage of the selection screen so a user simply has to enter an IP address and click on the "Run Nmap" button just underneath input panel for the IP address. This runs a bare bones Nmap scan.

I styled the button to match the aesthetic of the current buttons.

This video demonstrated the button's functionality and the underlying code:
https://youtu.be/pRz1zhAqEEk


TanvirRahmanTaiyeb (approved):
The quick scan button works well and blends nicely with the rest of the UI. I tested it and the scan triggered as expected. Also, the video was really helpful for understanding the update.

PR #1297, opened 2025-05-09
Add next step adviser

TanvirRahmanTaiyeb:
This update adds a new assistant that shows up after running a tool (e.g. Nmap) to help figure out what to do next. It uses GPT but only recommends tools that are actually available in the PT-GUI, based on what was found in the output.

- Created a new PentestGPT.tsx component that takes tool name and output and sends it to GPT.
- It suggests 1–3 follow-up tools (only GUI tools), with full reasoning and command-style examples.
- Hooked it into the Nmap interface to show how it works in a real use case.
- Uses getTools() from RouteWrapper, so the tool list is always synced.


jishnu5555 (approved):
I tested the PentestGPT feature with the Nmap tool, and everything is working as expected. PentestGPT effectively provides actionable next steps based on the Nmap output, tailoring its recommendations to the results and the tools available within the toolkit. This dynamic guidance significantly enhances usability. Overall, this is a valuable addition to the toolkit, and it would be great to see this feature extended to other tools as well.

PR #1310, opened 2025-05-12
Rtgen UI update

bshean1:
- Removed SaveOutputToTextFile and related code as it is not compatible with Rtgen.
- Changed some placeholder values as they were incorrect.
- Updated some input boxes to have both dropdown options and allow user input.
- Placed multiple input fields on the same line to reduce wasted space.


jishnu5555 (approved):
Great improvements on the UI front, the previous version occupied the entire screen, which made navigation and usability challenging. This PR resolves that issue effectively, resulting in a much cleaner and more user-friendly interface. The addition of the dropdown is a valuable enhancement that improves overall usability. The code cleanup looks solid, with no issues observed. I tested the tool locally, and it works perfectly while producing the expected output. Overall, this is a significant improvement to both the display and functionality of the RTgen tool. Thank you

PR #1312, opened 2025-05-13
Updated user guide

bshean1:
- Updated Photon's user guide to include the new advanced options.
- Included some more detail for already present options.


jishnu5555 (approved):
Great work on the Photon user guide update! The changes are thorough and clearly describe all the key features and advanced options within the toolkit. This update provides users with a solid understanding of the tool and the available configurations. Thanks for the contribution!

Issue #1000, opened 2024-11-18
[Bug]: Install process for dependencies - Loading screen update


GoldenBanksia:
### Name of Bug

Update additional tool dependency install loading screen 

### What Happened?

Automatic dependency installation process looks frozen or is confusing for end users. Seems to be stuck on the first debug message ‘apt-get update’, which can frustrate users who cancel this early – or makes it seem like the installation process does not work properly.  

Stuck on screen: 
Component Installation - Installation Progress - [Debug] Starting 'apt-get update' <datetime>

Makes it look like this is either frozen or not loading.

**Suggested fix**
Fix automatic install process by including a list of ‘Steps completed’ in addition to the current DEBUG output. Example of loading process language for end user to visually see how long the process will/or is taking: ...in progress..., ...done..., ...failed..., ...aborting...

Example: 
Component Installation
Installation process 
1. Apt-get-update...in progress 
2. Installing dependencies... 
3. Check command availability... 
 
_In addition_, add a **rendering icon, to signal it is not frozen**. _Then_ the current DEBUG output underneath this added information. 

### Steps to Reproduce

Tools -> GitLeaks -> try install dependency  

*This dependency will install correctly eventually for this tool, which is why it is a good example of the loading screen process

PR #1316, opened 2025-05-14
Bug #1000 - Loading Screen Update

Jatkin41:
- Improved initial download staging screen to improve the user experience and accessibility of Deakin Detonator Toolkit.
- Additionally improved the download in progress screen to contain a loading icon and verification that the download is in progress to address bug #1000 where it would appear the loading screen has frozen.

Attached images show the resulting changes.
Screenshot 2025-05-14 133337
Screenshot 2025-05-14 133351


Jatkin41:
ran prettier twice and it still doesn't work.. how fun

PR #1322, opened 2025-05-14
Fix: Add working config.yaml and update  cloudbrute.tsx for proper provider support

MrunalDas:
What This Pull Request Fixes

This pull request fixes the CloudBrute tool integration issue in the Deakin Detonator Toolkit (DDT) frontend UI, where running the tool via the graphical interface resulted in the following error:

Source detection failed as well use -c . error="no provider found in HTML"

The error originated from the fact that the DDT frontend did not correctly include the necessary provider configuration. The backend command that was constructed did not pass a valid --configFolder argument pointing to a properly structured YAML file, causing CloudBrute to fail.

Changes Made
	1.	Updated CloudBrute.tsx (Frontend Tool File)
	•	Added the --configFolder /etc/cloudbrute/config argument to the CloudBrute CLI call in the onSubmit function.
	•	This ensures that CloudBrute gets the proper configuration it expects when launched from the UI.
	2.	Corrected the structure of /etc/cloudbrute/config/config.yaml
	•	Reformatted the providers and environments keys to valid YAML array syntax.
	•	Removed malformed inline array that previously caused YAML unmarshalling errors like:
panic: yaml: unmarshal errors: line 2: cannot unmarshal !!str `-amazon...` into []string

Files Modified
	•	src/components/CloudBrute/CloudBrute.tsx
	•	/etc/cloudbrute/config/config.yaml

Only config.yaml was modified under config/. No changes were made to provider modules like google.yaml, microsoft.yaml, etc.

How It Was Tested
	•	After applying the fixes, the tool was tested directly in the Deakin Detonator Toolkit UI (GUI).
	•	The form was filled using:
	•	Domain: microsoft.com
	•	Keyword: test
	•	Wordlist: /usr/share/dirb/wordlists/common.txt
	•	After clicking Start CloudBrute, the scan executed successfully and showed:
                 microsoft detected
                 Initialized microsoft config

	•	No crash or YAML parsing errors occurred.
	•	The output verified that the config was parsed and the scan was working end-to-end.

The Screenshot of the tool working is shown below:

PHOTO-2025-05-15-00-06-53


jishnu5555 (approved):
The cloudbrute tool is working as expected, and the changes look great. I conducted testing with various inputs, and everything performed correctly. No other changes were made to the toolkit, so I'm approving the PR. Thanks for the work!

PR #1324, opened 2025-05-15
Added ADR for Feedback Feature

Bhavesharya07:
Created a new ADR entry describing the purpose, decision, and consequences of adding the Feedback feature.
ADR-0027.md


rng4302:
The feature seems appropriate and well explained so the PR is a good for merge, and the feature can be worked upon

TanvirRahmanTaiyeb (approved):
Thanks for submitting this ADR. The feedback feature is not currently aligned with the goals for this trimester, so the implementation will not go ahead for now. However, I’m approving this PR so the proposal is properly recorded in the documentation. This work can be revisited next trimester if it aligns with the goals at that time.

Issue #1199, opened 2025-04-11
[Improvement Request]: Organise Components into 'Tools' and 'AttackVectors' Folders


Oliver2049:
### Short name for suggested Improvement

Organise Components into 'Tools' and 'AttackVectors' Folders

### Priority

Medium

### Categories

Code Refactoring / Project Structure

### Description of Feature

Reorganise all components in the project by categorising them into Tools or AttackVectors folders. This is intended to improve clarity, organisation, and ease of maintenance for future contributors.

PR #1328, opened 2025-05-15
[Improvement] #1199 - Reorganising Tools and Attack Vectors

Oliver2049:
Now we have all tools and attack vectors in Tools and AttackVectors folders


TanvirRahmanTaiyeb:
This PR has been closed because while the reorganisation of tools and attack vectors is a good improvement, the older folders were not deleted. Keeping both the old and new structures causes conflicting files with the same names and increases the risk of confusion and maintenance issues.

It will be passed on to the next trimester’s students if it aligns with the upcoming goals.

PR #1329, opened 2025-05-15
docs(adr): ADR-0027 – Remove CrackMapExec from the Toolkit

yanand2:
### Summary

This pull request adds ADR-0027 documenting the formal decision to remove CrackMapExec (CME) from the Deakin Detonator Toolkit.


jishnu5555 (approved):
A great ADR, there were few minor issues with the ADR but fixed them and now it's looking good. Can confirm that the CrackMapExec tool is not maintained, thus the removal of the tool from the toolkit makes sense. A great initiative with the ADR, hence Approved. Thanks.

TanvirRahmanTaiyeb (approved):

PR #1333, opened 2025-05-15
Update MainLinks.tsx

Bhavesharya07:
Added feedback feature changes


TanvirRahmanTaiyeb:
Closing this PR as the implementation for the ADR was added, but the actual feature is not currently aligned with the goals for this trimester. The work has been recorded, but it won’t be moving forward at this stage. It can be picked up next trimester if it fits within the updated direction.

PR #1378, opened 2025-08-06
fix bug#1243 CloudBrute fix missing arguments

ataklaSec:
The Problem:

CloudBrute failed due to missing provider flag.

After provider flag was fixed, CloudBrute was causing a PT-GUI restart each time it was run.

CloudBrute it self has a output file it writes automatically to the root of the running application in this case it was the src-tauri root folder.

THE FIX:

Added provider flag to the tool code.
Added output flag to the tool code.
Added required GUI code for user entry.


Harmit3703 (approved):
Verified the fix locally. The missing --providers flag is correctly added, and the output flag integration works as expected. The tool now runs without crashing the GUI and produces the expected output. Everything looks good.

Approving the PR.

Issue #1400, opened 2025-08-11
[Bug]: Harvester Tool - "Google" source not supported despite being available in dropdown


chandupreethamm:
### Name of Bug

When attempting to run the Harvester Tool in Deakin Detonator Toolkit with the source set to "Google".

### Steps to Reproduce

1. Open Harvester tool in the tool kit 
2. Enter any valid domain 
3. Select source as Google
4. Click Start The Harvester

<img width="1493" height="799" alt="Image" src="https://github.com/user-attachments/assets/b8fd510f-f34b-4994-b686-a1090429bd22" />

fixed-by #1651

chandupreethamm:
> hello [@chandupreethamm](https://github.com/chandupreethamm) . The screenshots are not for the harvester tool. they are the same ones used for msfvenom missing dropdown payload bug#1383. could you please update with the correct screenshots so there is no confusion as to which tool you are referring. thank you

Hey @ataklaSec, updated with the original screenshot. Please do check it.

ataklaSec:
hello @chandupreethamm , i appreciate the reply and putting in the effort. thank you

typicalskip:
Fixed in PR #1651

PR #1414, opened 2025-08-23
Fix: Removed unsupported Google source from Harvester dropdown (#1400)

yanand2:
**Summary**
This PR resolves Issue #1400 by removing the unsupported "Google" source from the Harvester tool dropdown.
Previously, the Google option was visible in the UI but not supported by the backend, which caused confusion and runtime errors when selected.

**Changes**
Removed the "Google" option from the Harvester source dropdown in theharvester.tsx.
Added a code comment explaining that Google is currently unsupported, ensuring clarity for future developers.

**Notes**
This change ensures that only valid and supported sources are shown in the Harvester tool, improving usability and preventing users from selecting an invalid source.
Other errors (e.g., window.__TAURI_IPC__ is not a function) were observed when running with different sources, but these are separate issues not directly related to the removal of Google.

**Tested Outcome**
Verified locally:
Harvester tool loads without Google in the dropdown.
Confirmed that other sources (e.g., Bing, Shodan, LinkedIn) are still available.
The specific issue of Google appearing in the dropdown is resolved.
Other unrelated runtime errors remain and should be tracked separately.


Harmit3703:
Hello @yanand2, I see that References.tsx is also part of this commit, but it doesn’t look related to the fix for removing the unsupported Google source in the Harvester dropdown. Might be better to leave it out so the commit stays focused on just the issue being addressed.

SHELARARNAVI (approved):

Harsh965-glitch (approved):

typicalskip:
Obsolete PR, 0 files changed, associated issue was fix in PR #1651

PR #1436, opened 2025-09-02
bug#1418 update debian repo info

ataklaSec:
the problem:

README.md file had outdated Debian repository information that would stop the apt update if implemented as instructed.

the fix:

removed old buster repository information.
add bookworm repository information . 

NOTE: bookworm has now become old-stable under Debian release cycle as it will only have a couple of years before it will not be usable at best. this is a major issue as Debian will no longer support the libwebkit2gtk-4.0 after bookworm and will only have libwebkit2gtk-4.1 . this means tauri will have to be upgraded  from tauri v1 to tauri v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,8 +82,8 @@ To install Deakin Detonator Toolkit on Kali, you can follow the method below. Th
 4. Add the following as new lines to your APT sources list:
 
     ```bash
-    deb http://security.debian.org/debian-security buster/updates main
-    deb http://ftp.au.debian.org/debian buster main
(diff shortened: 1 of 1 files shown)


jishnu5555 (approved):
Hi @ataklaSec,

Thanks for your work on this PR.

The updates to the README.md file, replacing the outdated buster repositories with the currently supported bookworm repositories, are clear and necessary. This change will help users avoid issues during the apt update stage and ensures smoother setup of the toolkit.

Approving this PR. Thanks again for keeping the documentation up to date and for your attention to future compatibility concerns.

Issue #1388, opened 2025-08-09
[Bug]:  DDT Amass tool freezes indefinitely on invalid domain input with whitespace


tharinduLTM:
### Name of Bug

Amass tool hangs on invalid domain with whitespace

### What Happened?

When entering a domain containing a space (e.g :- exa mple.com) in the DDT Amass tool, the interface shows “The process is running” indefinitely and never returns output or an error message. This required manually cancelling the process.

<img width="602" height="343" alt="Image" src="https://github.com/user-attachments/assets/790d6803-e8e7-4e07-946c-0e5945a6cb67" />

<img width="602" height="342" alt="Image" src="https://github.com/user-attachments/assets/2ed12c34-da2f-4895-8a19-fcadbcdc60ad" />

### Steps to Reproduce

In “Enter the domain to scan”, type "exa mple.com". and click on "Sart Amass" button.

PR #1445, opened 2025-09-06
Fix for issue #1388 Amass missing input sanitization

MetricDunn:
Date fixed: 07/09/25
Tested by: James Dunn (MetricDunn)

Amass missing input validation: https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/issues/1388

Issue description:
Amass does not validate or sanitize user input, which caused DDT to hang when an improperly formatted domain was input.

Date reported: 09/08/2025
Reported by: MetricDunn

Fix summary:
Input validation was added to the user input form and now prevents Amass initiating a scan when there is improper input.

Files changed:
Deakin-Detonator-Toolkit/src/components/Amass/Amass.tsx

Commit reference:
https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/tree/issue-1388-amass-fix

Testing result:
I tested two domains a personal domain and one I was authorized to test and both succeeded and the the validation and sanitizing feature works. 

AMASS_FIXED_AND_FUNCTIONAL

AMASS_FUNCTIONAL_ERROR_CHECKING

### DIFF
--- a/src/components/Amass/Amass.tsx
+++ b/src/components/Amass/Amass.tsx
@@ -1,7 +1,6 @@
 import { Button, Stack, TextInput, Alert, Group, Text } from "@mantine/core";
 import { useForm } from "@mantine/form";
 import { useCallback, useState, useEffect, useRef } from "react";
-import { showNotification } from "@mantine/notifications";
 import { CommandHelper } from "../../utils/CommandHelper";
 import ConsoleWrapper from "../ConsoleWrapper/ConsoleWrapper";
 import { RenderComponent } from "../UserGuide/UserGuide";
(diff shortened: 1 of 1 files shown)


Harmit3703:
hello @MetricDunn , I tested your version and it works well — invalid domains now correctly show an error instead of hanging, and valid ones scan as expected.

One thought: what if we shift the regex validation directly to the input textbox level? That way the error is caught even before the scan function is called. This would:

- Prevent unnecessary function calls, saving system resources

- Give faster feedback to users while typing

- Further reduce the chance of Amass being invoked with invalid input

Your current fix works perfectly, but moving validation earlier could make the solution even more efficient.

MetricDunn:
@Harmit3703 I have implemented the changes that you requested. Please review once again and if you are happy, please merge this time.

PR #1452, opened 2025-09-08
bug_fix_#1408 Fixed Misleading Input Label and Missing Tutorial in Whois tool

KaniMahadevan:
PR Title
Fix Whois Input Field to Accept Only Domain or IP Address

Description
This PR updates the Whois tool to correctly handle user input and avoid confusion caused by the previous “Target URL” label.

What Changed

1. Renamed input field label from Target URL → Domain or IP address.
2. Updated user guide steps to clarify that only domain names or IP addresses should be entered (e.g., example.com, 93.184.216.34, [2001:db8::1]).
3. Validation check added in onSubmit: If a user enters http://, https://, or includes path/query/fragment characters, the tool blocks execution.
4. A clear message is displayed in the console: “Enter only a domain or IP address (no http/https, no paths). Examples: example.com, 93.184.216.34, [2001:db8::1]”
5. No other logic or toolkit structure was changed.

Why
Previously, the field label and guide text said “Target URL,” which misled users into pasting full URLs like https://example.com. This caused the whois command to fail with errors such as “No whois server is known for this kind of object.”

With this fix:

- The tool explicitly communicates what input is valid.
- Incorrect inputs are caught early with a friendly error message.
- User experience is clearer, preventing frustration.

Testing

Entered example.com → Whois lookup succeeds. (Image 1)
Image1

Entered 93.184.216.34 → Whois lookup succeeds. (Image 2)
Image2

Entered https://example.com → Blocked, error message displayed in console. (Image 3)
Image3

Entered example.com/test → Blocked, error message displayed in console. (Image 4)
Image4

### DIFF
--- a/src/components/Whois/Whois.tsx
+++ b/src/components/Whois/Whois.tsx
@@ -16,20 +16,29 @@ interface FormValuesType {
     targetURL: string;
 }
 
+/* --- Simple validators (no new deps) --- */
+const hasSchemeOrPath = (s: string) => /^\s*[a-zA-Z][a-zA-Z0-9+.-]*:\/\//.test(s) || /[\/?#]/.test(s);
+
+const ipv4Re = /^(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)(\.(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)){3}$/;
+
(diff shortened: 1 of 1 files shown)


Harmit3703 (approved):
Nice improvement here. The updated field label and validation make the expected input much clearer. I tested with both domains and IPs — everything works as described, and invalid inputs are properly blocked with a helpful message. This definitely prevents the old confusion around full URLs. Clear, user-friendly fix — approved.

PR #1457, opened 2025-09-10
issue1278 add shodanApi endpoints

ataklaSec:
the issue:

ShodanAPI was only limited to the host endpoint.
and enhancement by adding further endpoints would enhance the usability of the tool.

solution:

added count end point in shodkey.py file and the tool tsx file.
added search end point in shodkey.py file and the tool tsx file.
added facets end point in shodkey.py file and the tool tsx file.
added tokens end point in shodkey.py file and the tool tsx file.

### DIFF
--- a/src-tauri/exploits/shodkey.py
+++ b/src-tauri/exploits/shodkey.py
@@ -1,51 +1,155 @@
 import os, sys, getopt
 import shodan
+import json # Import json for pretty printing search results
 
 parameter_list = sys.argv[1:]
-options = "h:k:i:"
-long_options = ["Help", "Shodan Key", "IP Address"]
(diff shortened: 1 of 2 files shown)


Harmit3703 (approved):
Great job extending the ShodanAPI functionality. Adding the count, search, facets, and tokens endpoints really improves the usability of the tool beyond just the host endpoint.

I also noticed you updated the tutorial steps in Shodan-API-Tool.tsx while you were at it—good work including those, as they make the tool much clearer to use.

Overall, the implementation looks clean and consistent across both shodkey.py and Shodan-API-Tool.tsx. This is a solid enhancement and a valuable addition to the project.

Harmit3703 (approved):

PR #1470, opened 2025-09-13
Fix: Hydra form selects + validation for missing username/password

### LINKED ISSUE #1454: [Bug]: “No Username” / “No Password” input options not handled correctly
### Name of Bug

Input Options not handled

### What Happened?

In the Hydra component of the Deakin Detonator Toolkit, the form provides dropdown options for “No Username” and “No Password”. However, when these options are selected, the argument builder does not add any corresponding Hydra flags. Hydra requires at least one username (-l or -L) and one password (-p, -P, or -e) flag. Running the tool without these flags causes Hydra to error out immediately.

### Steps to Reproduce

Steps to Reproduce

Open the Hydra tool in the Toolkit.

Set Login settings to No Username.

Set Password settings to No Password.

Fill in other fields (Threads, Service, IP).

Click Crack.

Image

jonatjm06:
**Summary**

This PR fixes two issues with the Hydra feature:

1. The Login settings and Password settings selects were managed by two separate state sources, causing the conditional username/password fields to sometimes not render until the select was toggled. The selects are now controlled by useForm (single source-of-truth), which fixes the UI state and ensures the correct fields render on mount.
2. The form allowed submission that would run Hydra without any username or password flags (e.g. when users chose "No Username" / "No Password"). Since Hydra requires at least one username flag (-l or -L) and one password flag (-p, -P, or -e) and there are no Hydra flags for “no username / no password”, this PR enforces validation to prevent such submissions and shows user guidance on correct alternatives.

**What changed**
_Hydra.tsx:_

- Use form.getInputProps(...) for login/password selects to centralize state.
- Initialize sensible defaults for selects so inputs visible on mount.
- Mount-time sanity checks and validateSubmission() now prevent running Hydra without a username/password source.
- User-facing messages added to the console area explaining required inputs and alternative approaches (e.g., upload a file containing an empty line or use a placeholder like anonymous).

**Why**
- Prevents users from launching Hydra with invalid/missing required flags, which causes immediate failure and a confusing UX.
- Fixes a UI bug where inputs were hidden until toggling the select.

**Notes / future work**

If the team prefers automatic handling for empty username/password (e.g., auto-creating a temporary list file with an empty line), that would require back-end file upload or a temp-file endpoint. I can prepare a follow-up PR implementing automatic temp-file creation for the backend & frontend if desired.

### DIFF
--- a/src/components/Hydra/Hydra.tsx
+++ b/src/components/Hydra/Hydra.tsx
@@ -10,9 +10,6 @@ import { checkAllCommandsAvailability } from "../../utils/CommandAvailability";
 import InstallationModal from "../InstallationModal/InstallationModal";
 import { FilePicker } from "../FileHandler/FilePicker";
 
-/**
- * Represents the form values for the Hydra component.
- */
 interface FormValuesType {
     loginInputType: string;
     loginArgs: string;
@@ -26,7 +23,6 @@ interface FormValuesType {
     optionalConfig: string;
 }
 
-//Deals with the generatedfilepath unique identifier that is added at the end of a file
 const cleanFileName = (filePath: string): string => {
     const parts = filePath.split("_");
     const baseFileName = parts[0];
@@ -34,7 +30,6 @@ const cleanFileName = (filePath: string): string => {
 };
 
 const Hydra = () => {
-    // Component State Variables.
     const [loading, setLoading] = useState(false);
     const [output, setOutput] = useState("");
     const [pid, setPid] = useState("");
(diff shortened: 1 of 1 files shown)


Harmit3703 (approved):
I tested the changes and can confirm that the issue is fully resolved — Hydra no longer errors out when “No Username” / “No Password” is selected, and the UI state for the selects works correctly right from mount. The validation and user guidance you added make the workflow much clearer and prevent confusion.

Also, nice job on iterating and updating the PR with those extra tweaks in these last couple of weeks of the tri. Solid contribution and well-timed improvements — approving this one

Issue #1461, opened 2025-09-12
[Bug]: Some tools are missing Category values in Tools view


chandupreethamm:
### Name of Bug

Uncategorised tools not displayed properly in Tools list

### What Happened?

When browsing the Tools section, several tools appear with no Category assigned. This breaks the filter functionality because selecting a specific category hides these tools, making them undiscoverable.

Expected Behavior:
Every tool should have a valid Category, or there should be a default “Uncategorized” label shown and included in the filter list.

Actual Behavior:
Some tools display a blank Category. Filtering by category removes them from view entirely.

<img width="1469" height="152" alt="Image" src="https://github.com/user-attachments/assets/7b158863-e5e9-41fd-83c3-3f676cb6288a" />

<img width="1455" height="114" alt="Image" src="https://github.com/user-attachments/assets/13a4661c-117e-4bf9-92e3-e10edd09a9e7" />

<img width="1484" height="298" alt="Image" src="https://github.com/user-attachments/assets/82c736f7-c2eb-4614-93d0-5f21000bc7df" />

<img width="1591" height="333" alt="Image" src="https://github.com/user-attachments/assets/3e469f67-8d98-4dee-8a89-5f2bad8d16e2" />

### Steps to Reproduce

        1.	Open the Tools tab in PT-GUI.
	2.	Scroll through the list of available tools.
	3.	Observe that some tools (e.g., CloudBrute in my environment) have an empty Category field.
	4.	Use the Filter for a Category dropdown and select any category.
	5.	Notice that tools with no category assigned do not appear in the results and cannot be discovered unless
	       “All” is selected.

Suggested Fix
	•	Enforce Category as a required field for all tools.
	•	Backfill missing categories in the data.
	•	Provide a fallback option in the UI that shows “Uncategorized” if no category is assigned.
	•	Ensure the filter includes “Uncategorized” as a selectable value.

PR #1483, opened 2025-09-15
Automatic category creation to cover all tools - Bug fix for #1461

playBit1:
- The problem with the category dropdown was that it was static and outdated. 
- I generate an array of categories from the tools data as the solution. 
- Its wrapped in a useMemo hook to only regenerate the data if the tools’ data changes. This ensures that the dropdown will be UpToDate even if new categories are added 
- Category is mandatory for a tool so I didn’t add extra validation in this component

My comments for suggested fixes in the bug page:
• Enforce Category as a required field for all tools - **Category is already a required field for all tools**
• Backfill missing categories in the data - **All current tools have a category**
• Provide a fallback option in the UI that shows “Uncategorized” if no category is assigned & Ensure the filter includes “Uncategorized” as a selectable value. - **Unnecessary because all current tools have a category**

Before => After

- 
image

- 
image

- The performance cost was minimal when tested with the reactor profiler. Since I’m using useMemo, it will only run once when the app first starts, so performance cost could be disregarded.

Before:
image

After:
image

### DIFF
--- a/src/pages/Tools.tsx
+++ b/src/pages/Tools.tsx
@@ -1,23 +1,16 @@
 import { Stack, Table, Title, Select, useMantineColorScheme } from "@mantine/core";
-import { useState } from "react";
+import { useMemo, useState } from "react";
 import { getTools } from "../components/RouteWrapper";
 import ToolItem from "../components/ToolItem/ToolItem";
 
 const ToolsPage = () => {
-    const categories = [
-        "All",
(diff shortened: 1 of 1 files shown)


kani2003:
Thanks for adding Automatic category to cover all tools, though I noticed that your PR is failing style check. To fix it, can you please run the following command on the files you committed: "yarn prettier --write".

PR #1491, opened 2025-09-16
Fix MSFVenom Payload Dropdown Issues and Array Safety

deqny:
**Changes made:**

1. The payload dropdown now correctly shows the payloads for the default architecture on initial load, so users no longer see an empty dropdown when opening the tool.

2. Fixed the MAC x64 payload variables array to ensure it correctly matches the payloads, preventing potential array mismatches.

3. Added safety checks before accessing arrays in the payload selection handler. This prevents crashes if arrays are mismatched or if users rapidly change selections.

Testing done:

- Verified the payload dropdown shows correctly for the default architecture.

- Rapidly switched architectures and payloads — no crashes occurred.

- Generated payloads with different architectures and formats successfully.

### DIFF
--- a/src/components/msfvenom/msfvenom.tsx
+++ b/src/components/msfvenom/msfvenom.tsx
@@ -121,7 +121,7 @@ const payloadRequiredVariables = [
         ["LHOST", "LPORT"],
     ],
     //MAC x64 payload variables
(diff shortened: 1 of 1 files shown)


Harmit3703:
Thanks for the contribution @deqny. The fixes you’ve outlined for the payload dropdown, array safety, and crash prevention are valuable and well thought through. However, this PR currently has merge conflicts with main that need to be resolved before we can move forward. I’ll go ahead and close this one. Appreciate the effort you put into this!

PR #1500, opened 2025-09-17
fix(Apache-Apisix-2022-24112): wire PID and make Cancel send SIGTERM

Bhavesharya07:
### What
- Capture PID from spawned python process and store in state.
- Cancel sends 'kill -15 <pid>' when PID is known.
- Fallback to handle.kill() if PID is unavailable.

### Why
Previously Cancel sent 'kill -15' with an empty PID, so the python process kept running.

### Files
- src/components/CVE-2022-24112/CVE-2022-24112.tsx

### Testing
- Start exploit → click Cancel → process terminates; output shows termination soon after.
"


SHELARARNAVI (approved):

Harsh965-glitch (approved):

typicalskip:
0 files changed, PR is no longer relevant.

Issue #1446, opened 2025-09-06
[Bug]: WhatWeb tool: IP input fails, Aggression hangs, Plugins not detected, errors marked as success


sooraj126:
### Name of Bug

WhatWeb tool input, aggression, and plugin handling not working correctly

### What Happened?

I tested the WhatWeb tool with different inputs and scan options.

**1.  Domain input works fine e.g. google.com or https://www.google.com both return results normally. But entering an IP like 8.8.8.8 failed with:**

```
ERROR Opening: http://8.8.8.8 - execution expired
Process completed successfully.
```

So if IPs are not supported, the tool should clearly say “domain only” instead of failing. If they are supposed to work, then it needs fixing.

**2.   Changing Aggression Level from Default to Heavy caused the scan to hang indefinitely. After cancelling, it showed:**

```
Process was manually terminated.

```

**3.  Entering plugins like wordpress or apache gave error:**

```
No plugins selected, exiting.
Error: The following plugins were not found: wordpress
```
So the plugin input is not being passed properly.

**4.  In almost all error cases, the tool still ends with “Process completed successfully,” which is misleading.**

### **Expected Result**

- Domains should continue working as they do now.

- If IP input is not supported  it should show clear friendly message (“Only domain names supported”).

- If IPs should be supported   then it should be fixed  so they work properly.

- Aggression levels (Default, Heavy, etc.) should either run scans and return results, or stop after a reasonable timeout with a clear error instead of hanging forever.

- Plugin selection should correctly detect and run the specified plugins.

- Errors should clearly show what failed (invalid plugin, timeout, etc.), not mark failed runs as “completed successfully.”

### **Suggested Fix**

- Confirm whether IP scanning is supported.

           If yes then it should be fixed.

           If no then update tool to say “Domains only” in input field or user guide.

- Fix timeout issue for higher Aggression Levels. or stop after a set time with a clear error message (instead of hanging forever).

- Ensure plugin values entered in GUI are passed correctly to WhatWeb.

- Map raw errors to user friendly messages.

- Only show “Process completed successfully” when it actually completes successfully.

### Steps to Reproduce

1.  Open the WhatWeb tool.

<img width="1918" height="1007" alt="Image" src="https://github.com/user-attachments/assets/e9c33952-052d-4d06-a8ce-51aebbb75afd" />

2.  Enter google.com and press Run WhatWeb.

[...]

PR #1504, opened 2025-09-17
fix (whatweb) bug #1446

Bhavesharya07:
### What
- Build CLI args as separate tokens (fixes plugin passing).
- Add --timeout=30 default to avoid indefinite waits.
- Auto-cancel after 10 minutes (SIGTERM) to stop hangs.
- Map timeouts and plugin errors to clear messages; only show 'completed successfully' when truly successful.

### Files
- src/components/WhatWeb/WhatWeb.tsx

### Testing
- Domain and IP targets behave correctly; heavy scans no longer hang forever.
- Plugin names (e.g., 'wordpress,apache') are recognized and executed.
- Timeouts and invalid plugins show friendly messages.
"


typicalskip:
Superseded by #1648

Issue #1390, opened 2025-08-10
[Bug]: Arpaname tool dependency installation remains stuck and IP lookup hangs indefinitely


sooraj126:
### Name of Bug

Arpaname tool dependency installation remains stuck and IP lookup hangs indefinitely

### What Happened?

When launching the Arpaname tool, the dependency installation process never completes it remains stuck on “Required dependency is currently installing” indefinitely (waited up to 60 minutes). Closing the installation box allows the main page to load, but attempting to run an IP lookup causes the process to hang indefinitely with the message “Process is running, you can cancel below.” The cancel button is non functional, and the only way to exit is to reload the page by right clicking on the screen.

### Steps to Reproduce

1. Open the Arpaname tool.

2. On the dependency installation screen, click Install Components.

<img width="691" height="372" alt="Image" src="https://github.com/user-attachments/assets/c337dc47-3da1-4478-b5c8-d778cd4917ea" />

3. Enter the VM password when prompted.

<img width="691" height="284" alt="Image" src="https://github.com/user-attachments/assets/c8c0920d-2473-4046-9e39-07cfce5fcd6d" />

4. Wait for “Required dependency is currently installing” the process does not complete.

<img width="691" height="286" alt="Image" src="https://github.com/user-attachments/assets/cc55a447-528d-4496-9dc0-0a572b4d1102" />

5. Close the installation box using the X button.

6. On the main page, enter any IP address (tried multiple).

<img width="691" height="298" alt="Image" src="https://github.com/user-attachments/assets/94fe7647-6684-48f9-8f0c-43a787676acf" />

7. Click Lookup.

8. Observe that the “Process is running, you can cancel below” message remains indefinitely, and the cancel button does not work.

<img width="691" height="334" alt="Image" src="https://github.com/user-attachments/assets/8ca6f759-89ef-496a-a056-ccc133d9e79f" />

PR #1506, opened 2025-09-17
Change dependency to correct library: Bug #1390 fix

playBit1:
The tool in which the bug is present, uses the arpaname command from the bind9 kali tool. The main issue of dependency not being installed was fixed by changing the dependency to use the actual library name (bind9) instead of “arpaname”. 

Before:
image

486508361-00690ab6-a424-4813-aa0a-9e5b6842fc4c

After:
Screenshot 2025-09-18 053652

Screenshot 2025-09-18 053748

Screenshot 2025-09-18 053839

After changing the dependency the tool works successfully after the installation is complete


Harsh965-glitch (approved):

typicalskip:
Superseded by #1708

PR #1689, opened 2026-07-30
Fix ARPSpoof startup delay and Cancel Process hang (#1468)

### LINKED ISSUE #1468: [Performance]: ARPSpoof tool takes several minutes to start sending packets
**Description**
While testing the ARPSpoof module in the Deakin Detonator Toolkit, I noticed that the tool eventually works, but it consistently takes a long time (5-7 minutes) before forged ARP replies are actually observed on the network.

**Observed behavior**
DDT launches the arpspoof process, but no spoof packets are seen for several minutes.

Image

During this period, only legitimate ARP replies from the gateway (pfSense) appear in tcpdump.

Eventually, forged replies from Kali start appearing and the victim/gateway ARP tables update correctly.

Image

Running the same command manually on terminal works instantly.

**Expected behavior**
ARPSpoof should begin sending forged ARP replies immediately after the tool is started, consistent with manual usage of arpspoof.

**Suggested improvements**
Update the ARPSpoof command to use a single process with:
_/usr/sbin/arpspoof -r -t <victim> <gateway>_

**-r** ensures immediate bidirectional poisoning.

sruthi466:
Fixes #1468

### Problem
The ARPSpoof tool had two related issues:
1. No network interface was ever passed to the `arpspoof` command, leaving it to auto-detect — causing unreliable/slow startup.
2. Clicking Cancel Process while spoofing was running would hang indefinitely with no way to stop it short of killing the whole app.

### Root Cause
- `ArpSpoof.tsx` built its `arpspoof` arguments with only `-t <target1> <target2>`, never an `-i <interface>` flag.
- `arpspoof` runs as root via `pkexec`. The Cancel button's `handleCancel` (in `OverlayAndCancelButton.tsx`) tried to kill it with a plain `kill -2 <pid>` run as the normal user — which silently fails with a permissions error against a root-owned process, so the process was never actually signaled and the UI just waited forever for a `close` event that would never come.
- A secondary bug: the `if/else` around `pid2` was mismatched, so the "Failed to get process ID" error was incorrectly tied to `pid2` instead of the primary `pid`.

### Fix
- Added a required **Network interface** field to the ARPSpoof form, passed through as `-i <interface>` in both the gateway and target `arpspoof` commands.
- Updated `handleCancel` in `LoadingOverlayAndCancelButtonPkexec` to run `kill` via `pkexec` (`pkexec kill -2 <pid>`), so it has permission to signal the root-owned process.
- Fixed the `if/else` pairing so the PID-missing error only fires when the primary `pid` is actually missing.

### Testing
Tested on a local VM (VirtualBox NAT network, `10.0.2.3` / `10.0.2.2` as targets, `eth0` as interface):

**Before:** `pkexec arpspoof -t 10.0.2.3 10.0.2.2` (no interface flag). Spoofing worked, but Cancel Process hung indefinitely.

**After:** Interface field included; spoofing starts normally. Cancel Process now prompts `pkexec kill -2 <pid>` authentication, and both spoof processes report "Process completed successfully."

screenshot 1 before ARPSSPOOF

screenshot 2 before ARPSSPOOF

AFTER ..

screenshot 3 after ARPSSSPOOF

screenshot 4 after ARPSSPOOF


sruthi466:
Fixed the -i eth0 flag -tested, startup delay is gone, works instantly now.

While testing Cancel, found it doesn't actually kill the process (checked with ps aux). Dug in — the Cancel popup itself never shows up on screen for ArpSpoof, even though it works fine on other tools like Netcat. So it's not really my kill-command logic that's broken, it's that the button never appears in the first place. Seems specific to ArpSpoof, maybe because it runs two commands at once (gateway + target) instead of one.

-i fix is done and tested. Cancel needs a separate fix for the popup not showing — can take that up next if needed.

Harsh965-glitch (approved):

Issue #1520, opened 2025-09-21
[Bug]: Dependency Installation & Authentication Flow Problems


hades7200:
### Name of Bug

Gitleaks

### What Happened?

During installation of GitLeaks via the Deakin Detonator Toolkit, dependency setup fails due to missing GPG keys and repeated authentication prompts. This prevents GitLeaks from being installed successfully and confuses users.

**Actual Result:**

apt-get update fails with NO_PUBKEY error.

User must enter password multiple times.

Installer freezes with no clear recovery instructions.

**Expected Result:**

GitLeaks installs smoothly with dependencies.

Only one authentication prompt should appear.

Clear error handling for missing GPG keys (e.g., instructions or automatic key import).

**Suggested Fix:**

Bundle GitLeaks and required dependencies into installer OR provide automated script to resolve missing keys.

Consolidate authentication into a single sudo request.

Improve installation feedback with progress logs and recovery guidance.

### Steps to Reproduce

Steps to Reproduce:

Open Deakin Detonator Toolkit → Tools → GitLeaks.

Start the installation process.

Enter sudo password when prompted.

Observe multiple authentication prompts and eventual GPG error.

<img width="1918" height="901" alt="Image" src="https://github.com/user-attachments/assets/66904a11-471c-4849-b097-7854a7efef20" />
<img width="1909" height="915" alt="Image" src="https://github.com/user-attachments/assets/1a56c675-8b3e-4171-b3a7-8a04a6704087" />
<img width="1919" height="891" alt="Image" src="https://github.com/user-attachments/assets/8989717e-e926-4f65-b839-298c6abac313" />

<img width="1919" height="1042" alt="Image" src="https://github.com/user-attachments/assets/a0a40c78-43e6-49cb-9f2f-54b395e3ce83" />

PR #1692, opened 2026-08-02
Fix GitLeaks installation failure (#1520)

Suthir77:
## Summary
Fixes the GitLeaks installation failure described in task #1520, where installation failed due to missing GPG keys and repeated authentication prompts.

## Root Cause
The previously used GitLeaks install script URL (`https://raw.githubusercontent.com/gitleaks/gitleaks/master/scripts/install.sh`) returns a 404 and is no longer valid, causing the installation to fail immediately.

## Changes
- Added `scripts/install-gitleaks.sh`, a new installation script that:
  - Downloads the GitLeaks binary directly from the official GitHub Releases page instead of relying on a fragile install-script URL
  - Consolidates `sudo` authentication into a single upfront `sudo -v` check (with a background keep-alive loop) to avoid repeated password prompts during installation
  - Verifies the installation was successful and reports a clear success/failure message

## Testing
Ran the script on a Kali Linux VM and confirmed GitLeaks installed successfully (version 8.18.4) with no GPG errors and only one authentication prompt.

## Checklist
- [x] Reproduced the installation failure
- [x] Identified root cause (broken install script URL, not literally a GPG key issue as originally suspected)
- [x] Fixed installation to complete successfully
- [x] Reduced repeated authentication prompts
- [x] Verified successful installation


typicalskip:
This PR is superseded by #1734

PR #1708, opened 2026-08-13
fix(arpaname): prevent install and lookup hangs (#1390)

### LINKED ISSUE #1390: [Bug]: Arpaname tool dependency installation remains stuck and IP lookup hangs indefinitely

sahil296:
Summary

Fixes #1390 where Arpaname could get stuck during dependency installation and lookups could hang indefinitely.

What changed

* Fixed the dependency check so it looks for the arpaname command but installs the correct package, bind9
* Fixed the install flow so it properly handles success, failure, cancellation, and timeout
* Fixed lookup process handling so it waits for the real process to finish
* Made Cancel reliably stop the active process
* Prevented multiple install or lookup processes from starting at the same time
* Added cleanup for timers and process references so the UI does not remain stuck
* Rechecks for arpaname after installation
* Removed the duplicated validation error message

Tested:

* bind9 installation through DDT
* cancelling and retrying the installation
* IPv4 lookup
* IPv6 lookup
* repeated lookups
* process cancellation
* timeout/error handling
* reopening the tool after installation

Files changed

* src/components/Arpaname/arpaname.tsx
* src/components/Arpaname/arpanameProcess.ts
* src/components/Arpaname/ArpanameInstallationModal.tsx

Closes #1390


typicalskip:
Please make sure you run `corepack yarn test --run` and `corepack yarn build`.
Build check and test check failed, arpanameProcess.ts cannot locate "@tauri-apps/api/shell", check new Tauri 2 import for shell.

typicalskip (approved):
Looks good

PR #1713, opened 2026-08-16
Add target reachability check and TLS warning to BED tool (#1425)

### LINKED ISSUE #1425: [Bug]: BED Tool Error Handling and Service Validation
### Name of Bug

Error Handling and Service Validation

### What Happened?

**Problems Identified:**

Error messages are unclear – Right now, if a scan fails it just says “connection refused”. It doesn’t tell the user which host or port was tried. For example, it should say something like “Connection refused at 127.0.0.1:80” so people know exactly what failed.

TLS support isn’t explained – BED is an older tool and doesn’t support TLS. This means scans against common services like Gmail SMTP or IMAP which require TLS will always fail. The UI doesn’t make this obvious, so users might waste time trying.

Scans run even if no service is there – The tool will happily start a scan even when nothing is listening on the target. This always fails and just confuses users.

Image
Image
Image

### Steps to Reproduce

**Proposed Solutions:**

Make error messages clearer by showing which host and port failed.

Add a small warning or tooltip in the UI explaining that TLS isn’t supported, and that users should test against plain local services instead also removing any TLS options.

Do a quick check before scanning to confirm the target service is actually running. If it’s not reachable, let the user know right away.

sruthi466:
## Summary
Fixes #1425 by adding target reachability validation and a TLS support warning to the BED Tool.

## Changes
- Added a `check_target_reachable` Tauri command that performs a TCP connection check (3s timeout) against the target host/port before BED starts.
- If the check fails, BED does not start. The loading state stops and a clear message is shown (e.g. "Connection refused at 127.0.0.1:9999", "Connection to 10.255.255.1:80 timed out"), and the user can retry immediately.
- Added a visible warning banner for SMTP, POP, and IMAP explaining that BED does not support TLS and should only be tested against plain/local/lab services.

## Testing
Manually tested all required cases:
- Reachable service (local Python HTTP server) - check passes, BED launches and produces real output
- Connection timeout (unroutable IP) -clear timeout message shown
- Closed port -clear "Connection refused" message shown
- TLS warning -banner appears when SMTP/POP/IMAP is selected as the plugin


Harsh965-glitch (approved):

typicalskip:
Please clear branch conflicts

Harsh965-glitch (approved):

Harsh965-glitch:
hey so this pull request fails the checks that are test and build checks

Rowy5000 (approved):

typicalskip (approved):
Looks good

PR #1746, opened 2026-09-01
Fix/1694 crunch dependency prompt

### LINKED ISSUE #1694: [Bug]: Dependency installation prompt appears every time the tool is opened
### Name of Bug

crunch -Dependency installation prompt 

### What Happened?

**Dependency installation prompt appears every time the tool is opened** 
Whenever the Crunch tool is opened in the Deakin Detonator Toolkit (DDT), a dependency 
installation window appears. The installation process displays the message "Unable to locate 
package Crunch", but the window remains open even after showing "Process completed 
successfully." The user has to manually close the installation window before accessing the tool.
Although the Crunch tool works correctly after manually closing the installation window, the 
behaviour may confuse users into thinking that the installation has failed or that the tool is 
unavailable. 

### Steps to Reproduce

The installation window should automatically close once the dependency check is complete if 
the Crunch tool is already available on the system. Alternatively, DDT should display a clear 
message indicating that the tool is already installed and ready to use, allowing the user to 
proceed without manual intervention.

SachinKumar3303:
Fixes #1694

The Crunch component was checking for the dependency using "Crunch" instead of the actual Linux command name "crunch".

Since the command availability check uses the `which` command, the uppercase dependency name was not detected even though Crunch was already installed. This caused the dependency installation prompt to appear every time the tool was opened.

Changes:
- Updated the Crunch dependency name from "Crunch" to "crunch".
- Synced the branch with the latest main branch and resolved the Crunch merge conflict.
- Verified that `/usr/bin/crunch` is detected correctly.
- Tested the fix in the Tauri application.
- Confirmed that the installation prompt no longer appears when reopening Crunch.
- Confirmed that Crunch still generates wordlists successfully.


typicalskip (approved):
minimal changes, looks good

PR #1703, opened 2026-08-09
Add foremost command configuration to tauri.conf

dc19837:
#1547

Fixes #1709

PR #1763, opened 2026-09-07
Update tauri configuration with shell command allowlist

dc19837:
Added shell command allowlist for various tools in tauri configuration. Ran corepack yarn add @tauri-apps/plugin-fs and corepack yarn build which passed.

### DIFF
--- a/src-tauri/tauri.conf.json
+++ b/src-tauri/tauri.conf.json
@@ -32,6 +32,493 @@
       "providerShortName": null,
       "signingIdentity": null
     },
+    "tauri": {
+        "allowlist": {
+            "shell": {
+                "open": true,
+                "execute": true,
+                "sidecar": true,
+                "scope": [
+                    {
+                        "name": "nmap",
+                        "cmd": "/usr/bin/nmap",
+                        "args": true
+                    },
+                    {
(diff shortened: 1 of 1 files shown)


typicalskip:
Hi, is there a reason this has been added directly to the tauri.conf.json.  According to the tauri 2 documentation (https://v2.tauri.app/plugin/shell/), permissions are set in the capabilities file.  For this project see src-tauri/capabilities/migrated.json, there is an allow list for both shell:allow-execute and shell:allow-spawn and a blanket shell:allow-open.  Possibly tools are not included on these lists.

dc19837:
I added the entry to tauri.conf.json because Foremost was not present in migrated.json, and during testing this change allowed Foremost to execute correctly through DDT and generate the expected output.

dc19837:
Addressed the review feedback by implementing the Foremost permission through the Tauri capabilities system in src-tauri/capabilities/migrated.json. Added Foremost to both shell:allow-execute and shell:allow-spawn, consistent with the existing tool configuration pattern. This supersedes the previous tauri.conf.json approach used in #1703.

Rowy5000 (approved):

PR #1765, opened 2026-09-08
Fix WhatWeb input file target processing

### LINKED ISSUE #1723: [Bug]: WhatWeb input-file option does not process valid targets
### Name of Bug

Valid input file returns no targets selected

### What Happened?

WhatWeb does not successfully process a valid target supplied through the Input File field.

### Expected behaviour

When a valid input file contains a supported target such as:

`http://127.0.0.1:8000`

WhatWeb should read the target from the file and perform the scan.

### Actual behaviour

The GUI accepts the input file and allows the user to continue through the workflow. However, the resulting WhatWeb execution reports:

`No targets selected`

The process then completes without scanning the target.

This occurs even though the selected text file contains a valid local HTTP target.

### Steps to Reproduce

1. Create a text file containing:
   `http://127.0.0.1:8000`
2. Open WhatWeb.
3. Leave Target URL or IP blank.
4. Enter/select the text file in the Input File field.
5. Continue using default scan options.
6. Run WhatWeb.
7. Observe that the process reports “No targets selected.”
8. Save or open the resulting output and confirm that no target was scanned.

Meaad2026:
Fixes #1723

This PR fixes WhatWeb input file target processing by correctly passing the selected input file to WhatWeb using the -i argument. It also prevents an empty target value from being added when the user is using an input file only.

As an additional usability improvement, a Browse button was added so users can select a .txt input file directly instead of manually entering the full file path.

Testing confirmed that WhatWeb successfully reads the selected file and scans the target, returning a 200 OK result.

### DIFF
--- a/src/components/WhatWeb/WhatWeb.tsx
+++ b/src/components/WhatWeb/WhatWeb.tsx
@@ -22,6 +22,7 @@ import { LoadingOverlayAndCancelButton } from "../OverlayAndCancelButton/Overlay
 import { checkAllCommandsAvailability } from "../../utils/CommandAvailability";
 import InstallationModal from "../InstallationModal/InstallationModal";
 import { RenderComponent } from "../UserGuide/UserGuide";
+import { open } from "@tauri-apps/plugin-dialog";
 
 /**
  * Represents the form values for the WhatWeb component.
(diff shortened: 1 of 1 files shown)


typicalskip (approved):
Looks good

PR #1766, opened 2026-09-08
Fix #1712: Add worker/socket input validation to GoldenEye tool

### LINKED ISSUE #1712: [Bug]: GoldenEye accepts negative worker values and reports successful completion
### Name of Bug

GoldenEye accepts negative concurrent worker values

### What Happened?

The GoldenEye configuration form accepts -1 as the number of concurrent workers and passes the value directly to the Python script:

python3 exploits/Goldeneye/goldeneye.py http://127.0.0.1:8000 -w -1 -s 1 -m get

GoldenEye then displays:

Hitting webserver in mode 'get' with -1 workers running 1 connections each.

Because the worker value is negative, no workers are created and no requests are sent. Despite this, the output displays:

Shutting down GoldenEye

Process completed successfully.

Expected behaviour: DDT should reject zero, negative, decimal and non-numeric worker values before launching GoldenEye. The field should require a positive integer of at least 1 and provide a clear validation message.

Actual behaviour: The negative value is accepted, the process launches with -1 workers, performs no test and misleadingly reports successful completion.

The same positive-integer validation should also be applied to the concurrent sockets field. Validation should be enforced in both the DDT interface and the GoldenEye backend so that invalid values cannot be passed directly through another execution method.

### Steps to Reproduce

1. Open the Deakin Detonator Toolkit.

2. Navigate to the GoldenEye configuration screen.

3. Enter http://127.0.0.1:8000 as the controlled local target.

4. Leave the user-agent field blank.

5. Enter -1 for the number of concurrent workers.

6. Enter 1 for the number of concurrent sockets.

7. Select the GET method.

8. Click Launch DoS Attack.

9. Observe that DDT accepts the negative value and generates a command containing -w -1.

10. Observe that GoldenEye starts with -1 workers, sends no requests and reports Process completed successfully.

sruthi466:
Fixes #1712

Adds validation for the Workers/Sockets fields in GoldenEye so invalid input can't be submitted, silently dropped, or bypass the UI.

Changes:
- Frontend: added `validate` rules requiring Workers/Sockets to be non-empty, digits-only, ≥ 1. Also fixed a truthy-check bug where `0` was silently dropped from the command instead of being sent (`0` is falsy in JS).
- Backend: added matching `isdigit()` + `< 1` checks in `goldeneye.py` so invalid values are rejected even if the script is called directly.

Tested: 
empty, `0`, negative, and decimal/non-numeric input are all rejected with clear messages (UI + backend). Invalid runs exit non-zero, so no false "Process completed successfully." Valid input (workers=5, sockets=20) still launches correctly against a local test target.

Note: found a pre-existing, unrelated bug where GoldenEye's script path resolves incorrectly in dev mode (`src-tauri/src-tauri/...`). Out of scope for this PR - happy to file it separately.

## Screenshots
Screenshot 2026-09-08 231315

Screenshot 2026-09-08 231352

Screenshot 2026-09-08 231431

Screenshot 2026-09-08 231504

Screenshot 2026-09-08 232609


Rowy5000 (approved):

Rowy5000 (approved):

typicalskip:
Please make sure you run `corepack yarn style`, `corepack yarn prettier --write .`.

typicalskip (approved):
Looks good

PR #1779, opened 2026-09-11
Fix Metagoofil zero-result and rate-limit handling

### LINKED ISSUE #1635: [Bug]: Metagoofil Returns 0 Results for PDF Searches Across Multiple Domains
### Name of Bug

Metagoofil returns 0 results

### What Happened?

While testing the Metagoofil tool within the Deakin Detonator Toolkit (DDT), the tool consistently returned 0 PDF results across multiple domains, even when valid targets were provided.

The same behaviour was reproduced directly in the native Kali Linux CLI, indicating that the issue is not caused by DDT integration but is likely related to limitations of the Metagoofil tool itself.

During testing, it was observed that Metagoofil is no longer actively maintained and modern search engines such as Google heavily restrict automated scraping requests, which affects the tool’s ability to retrieve indexed documents.

### Steps to Reproduce

1. Open the Metagoofil module in DDT
2. Enter a valid domain (e.g., apple.com)
3. Select PDF as the target file type
4. Start the scan
5. Observe that the tool returns:
   “0 PDF files found”

The same command was also tested directly in Kali Linux CLI with identical results.

Image
Image
Image

sachstephperera:
## Summary

This PR fixes GitHub Issue #1635 related to Metagoofil returning 0 results for PDF searches.

The Metagoofil component now handles zero-result and search-provider restriction cases more clearly instead of making them look like generic execution failures.

## Changes

- Added clearer handling for Metagoofil zero-result responses.
- Added handling for HTTP 429 / search-provider rate-limit messages.
- Normalized domain input before running the command.
- Normalized file-type input, for example `.PDF` to `pdf`.
- Added clearer output messages in the DDT interface.
- Improved process completion handling.
- Added a safer delay between automated search requests.

## Testing

Testing was completed in Kali Linux using the Deakin Detonator Toolkit desktop application.

Test case:

- Domain: `apple.com`
- File type: `pdf`

The command executed successfully:

`metagoofil -d apple.com -t pdf -e 60`

Metagoofil returned:

`Results: 0 .pdf files found`

DDT then correctly displayed:

- No matching files were returned by Metagoofil.
- Modern search engines may restrict automated searches.
- Zero results do not necessarily mean the target domain contains no matching files.
- Process completed successfully.

The same zero-result behaviour was also reproduced directly from the Kali Linux command line, confirming that the underlying search behaviour is from Metagoofil/search-provider limitations rather than the DDT interface.

## Issue

Closes #1635


typicalskip:
Please make sure you run `corepack yarn style`, `corepack yarn prettier --write .`

sachstephperera:
Done, I ran corepack yarn style and corepack yarn prettier --write ..
Both completed successfully, and there were no source formatting changes to
commit. Thanks!

typicalskip:
it is still showing prettier fails.  Prettier was updated to a new version on main, you may need to either update your local prettier or pull and merge your fork with the main (if you do this make sure you have a copy of the file you changed so you can quickly restore what you have done)
image
image

PR #1793, opened 2026-09-13
Fix #1715: surface AskChatGPT failures instead of failing silently

### LINKED ISSUE #1715: [Improvement Request]: Ask ChatGPT button is unresponsive after a successful Sherlock search

Make the Ask ChatGPT button responsive

### Priority

Medium

### Categories

UI, UX

### Description of Feature

Expected Behaviour: The Ask ChatGPT button should respond after the search completes, such as by generating an AI-assisted analysis of the search results or displaying appropriate feedback if the feature is unavailable.

Actual output: When Sherlock successfully completes a search, there is no visible response when we click the Ask ChatGPT button. The program doesn't produce an error message, open a new window, display an analysis, or display a loading indication.

Steps to reproduce:
1.	Open DDT 
2.	Navigate to Tools. 
3.	Open the Sherlock Tool. 
4.	Leave Advanced Mode turned Off. 
5.	In the Username field enter a valid username.
6.	Click “Start Searching!”. 
7.	After the results are successfully displayed, click “Ask ChatGpt” button.
8.	Observe the output.

BiOneIsDaBest:
Closes #1715

## Problem

The "Ask ChatGPT" button in Sherlock appears to do nothing when clicked.

It is not actually unresponsive. The handler runs, but the request fails in under a millisecond because `VITE_OPENAI_API_KEY` is not configured (only `.env.example` exists in the repository). Three issues combine to hide this from the user:

1. The failure is instantaneous, so React batches the state updates and the loading state never renders.
2. The inner `catch` returned the error as a string instead of propagating it, so it was treated as a successful response.
3. That string was written into `chatGPTResponse` and rendered at the bottom of the page under the "ChatGPT Response" heading, directly beneath a disclaimer stating that AI responses may be inaccurate — so a configuration fault was presented as if it were an AI answer.

## Scope

This is not Sherlock-specific. The button comes from the shared `src/components/AskChatGPT/AskChatGPT.tsx`, which is imported by ten tools: Sherlock, Nmap, Dig, Traceroute, wafw00f, Subjack, Enum4Linux, NSLookup, Amass and Arpaname. The fix is made in the shared component, so all ten benefit.

## Changes

Single file, props interface unchanged — no consuming tool needs modification.

- Errors go into a dedicated `error` state rendered in a Mantine `Alert` directly beneath the button, not into `setChatGPTResponse`
- Removed the inner `try/catch` around the fetch so errors propagate to one handler
- Distinct messages for HTTP 401 (invalid key) and 429 (rate limit)
- The missing-key message now names the variable, says where to set it, and points to `.env.example`
- Added Mantine's `loading` prop, plus a hint explaining the disabled state while no tool output exists
- Previous response is cleared when a new request begins
- Caught error typed safely with `instanceof Error`

## Testing

Verified through the Vite dev server (`yarn dev`) on Sherlock:

- Before a search: button disabled, precondition now stated below it
- After a search: clicking produces a visible alert beneath the button explaining the configuration problem and how to fix it
- The "ChatGPT Response" panel no longer appears on failure
- `npx tsc --noEmit` reports no errors in the modified file; Prettier passes

GUI testing inside the packaged Tauri app was not possible — `src-tauri/tauri.conf.json` on `main` currently has a JSON syntax error at line 522 (leftover Tauri v1 config from the v2 migration), so `yarn tauri dev` fails to start. This is a project-level issue unrelated to this PR and has been raised with the team lead separately.

## Note for the team

`src/components/AskCohere/AskCohere.tsx:87` has the same pattern — the error is written into the state used for successful responses. Worth a follow-up issue.


typicalskip:
Please make sure you run `corepack yarn style`, `corepack yarn style:fix`

PR #1795, opened 2026-09-13
Update upload icon source in Rainbowcrack component

Sluubby:
Quick fix to mend the broken icon for the path picker

The icon of the file handler has been updated. The image path is reused for Rainbowcrack .rt path picker.


typicalskip (approved):
looks good

PR #1798, opened 2026-09-14
#1683 Feature Improvement: Allow BEDTool navigation while a scan is running

### LINKED ISSUE #1683: [Improvement Request]: Allow BEDTool navigation while a scan is running

Background scan execution

### Priority

Medium

### Categories

UI, UX, Enhancement, Performance

### Description of Feature

BEDTool should allow scans to run in the background without locking the entire interface.

At present, starting a scan opens a process overlay that prevents navigation through the rest of BEDTool. While the scan is active, the user can only wait for it to finish or select “Cancel Process.”

Image

This is particularly limiting during long scans. In testing, a complete HTTP scan took approximately two to three hours in the available virtual-machine environment. During that time, the User Guide, Tutorial, configuration controls and other toolkit areas could not be accessed.

The improved version should allow users to:

- minimise or close the active scan panel without cancelling the scan;
- navigate to the User Guide and Tutorial;
- access other non-conflicting parts of the toolkit;
- view elapsed time and current scan status;
- return to the scan output at any time;
- receive a notification when the scan finishes;
- cancel the scan when required.

The scan should continue safely in the background while preventing only actions that would interfere with the active process.

This improvement would make BEDTool more practical during long-running tests and reduce unnecessary waiting.

Sluubby:
By clicking outside or pressing ESC, users can have BEDTool scans to run in the background, freeing up navigation to the User Guide, Tutorial, and other DDT areas while a scan is in progress. 
Configuration fields remain editable during an active scan, though a new scan cannot be submitted until the current one finishes or is cancelled. Scan progress (console output and completion results) now persists beyond component unmounts, so navigating away from BEDTool and back no longer resets an in-progress or loses a completed scan.

Fixes #1683 

Note: During testing, I found that the summary board is programmed for a completed, cancelled scan but not disconnected internet connection. And, clearing the output does not reset/collapse the summary.


typicalskip (approved):
looks good

PR #1806, opened 2026-09-16
fix(cve-2022-26134): stabilize repeated runs, validate input, and clarify results (#1631)

### LINKED ISSUE #1631: [Improvement Request]: CVE-2022-26134 Vulnerability Tool Stability

CVE Tool Stability

### Priority

Medium

### Categories

Reliability

### Description of Feature

This tool works properly on the first run but causes the application to restart when executed multiple times. It also does not handle invalid input properly and shows technical errors instead of clear messages. In addition, the output does not clearly indicate whether the target is vulnerable. Improvements are needed for stability, input validation, and output clarity. I have linked my testing report below.

[Testing.pdf](https://github.com/user-attachments/files/27330348/Testing.pdf)

sahil296:
Summary

Fixes [#1631](https://github.com/Hardhat-Enterprises/Deakin-Detonator-Toolkit/issues/1631). The CVE-2022-26134 (Confluence OGNL injection) tool had three reliability problems reported in testing:

Stability — it worked once, then restarted the whole desktop app when run again.
Input handling — invalid input produced raw Python tracebacks instead of guidance.
Output clarity — results never plainly stated whether a target was vulnerable.

This PR addresses all three, following the same approach already merged for the Sherlock "repeat search" fix.

**Root cause of the restart**

The exploit script always wrote a results file into the current working directory: the -o option defaulted to exploit_result.txt, so if optionsOpt.output: was always true. Inside the desktop app that directory is watched by Tauri, so every run wrote a file into the watched tree and triggered an app restart on repeated runs. This is the same class of bug fixed for Sherlock (output written into the watched project directory).

**Changes**

**Stability**

The UI now resolves appCacheDir() and passes -o <cache>/CVE-2022-26134-results.txt, so the script's file write lands outside the watched project directory.
The script no longer writes any file by default (the -o default is now empty). It only writes when a path is explicitly provided, and it creates the parent directory first so the write is reliable.

**Input handling** 

Frontend validation before launch: a target is required (URL in single mode, file path in multi-target mode), and the timeout, if provided, must be a positive whole number.
The Python script exits cleanly with clear messages for: no target supplied, target file missing/empty/unreadable, non-positive timeout, and a missing requests dependency.

**Output clarity**

Each target now reports an explicit [+] VULNERABLE or [-] NOT VULNERABLE line (including the HTTP status, or the connection/timeout/error reason), followed by a Scan complete. Vulnerable targets found: N summary.
The Timeout value is now actually applied to each HTTP request (previously it was parsed but never used), and the UI field gained a description and placeholder.

**Maintainability**

Argument construction and validation were extracted into a pure, tested cve202226134Args module.
Removed the globals().update(locals()) hacks, the unused recursion-limit call, and brittle open() usage; worker parameters are now passed explicitly via functools.partial instead of globals.


typicalskip:
looks good, but did you mean to upload your test files?

PR #1809, opened 2026-09-17
fix/sherlock-delimiter-1807

### LINKED ISSUE #1807: [Bug]: Sherlock Username Input Can Be Interpreted as CLI Options
### Name of Bug

Sherlock Argument Input Validation

## Description

The Sherlock username field currently allows input to be interpreted as Sherlock command-line options rather than strictly as username data. entering a valid option into the Username field caused Sherlock to interpret the value as a CLI option

## Expected Behaviour

Any value entered through the Username field should be treated strictly as positional username data

## Actual Behaviour

Username input can be interpreted as Sherlock CLI options meaning other Sherlock options may potentially be interpreted in the same way when supplied through the Username field

## Steps to Reproduce

1. Open the Sherlock tool
2. Leave Advanced Mode disabled
3. Enter an option in the following into the Username field, For example:

   `--version`

6. Click "Start Searching!"
7. Observe the output

The resulting command is equivalent to:

`$ sherlock --timeout 60 --version`

Sherlock interprets `--version` as an option and returns its version information instead of performing a username search

Image

Suthir77:
Add -- delimiter before username args to prevent CLI flag injection

Closes #1807


typicalskip (approved):
looks good

Issue #1725, opened 2026-08-19
[Improvement Request]: Allow WhatWeb navigation while a scan is running


Llewellyn-Jordaan:
### Short name for suggested Improvement

Background scan execution

### Priority

Medium

### Categories

UI, UX, Enhancement, Performance

### Description of Feature

WhatWeb should allow the user to access non-conflicting areas of the interface while a scan is running.

At present, starting a scan prevents navigation through the tool until the process finishes. The only available action is to cancel the active scan.

Although WhatWeb scans were relatively fast during testing, this still limits usability and may become more significant when scanning slower or more complex targets.

The improved version should allow users to:

- minimise or background an active scan;
- access the User Guide and Tutorial;
- navigate to other non-conflicting toolkit areas;
- view elapsed time and scan status;
- return to the active scan at any time;
- cancel the process when required.

This would make WhatWeb more flexible and improve the user experience during longer scans.

typicalskip:
Fixed by #1811

PR #1811, opened 2026-09-19
Allow WhatWeb navigation while scan is running

joabjunxianglow:
Implements the enhancement request as described by #1725.

Changes: 
-  With inspiration from #1798, similar changes have been made to WhatWeb.tsx. 

Notes: 
- Once the scan has terminated, it takes a moment for the WhatWeb interface to reflect that termination where another scan cannot begin.


typicalskip (approved):
looks good

